Issue #21, opened 2021-08-19
Path should be regex escaped


erwin-faceit:
The help states that arguments will be used to filter test files, and "each argument is treated as a regular expression". At the moment the path is passed as argument, which (on Windows at least) is not a valid regular expression. This results in 

```
=================
 no tests found.
=================
```

I think the path passed on the lines below (line 109 specific) should be escaped to be regular expression safe. In my case, if I replace all backslashes with double backslashes, the test is executed correctly.

https://github.com/microsoft/playwright-vscode/blob/959031d8e6ae5764ec8cdb088bf671800ab5e8b6/src/playwrightTest.ts#L106-L110

I think the same goes for https://github.com/microsoft/playwright-vscode/blob/959031d8e6ae5764ec8cdb088bf671800ab5e8b6/src/playwrightTest.ts#L62

Finally, on line 47 a shortcut is used `replaceAll('\\', '\\\\')`, but I think this should be a method which escapes all regex characters instead.

https://github.com/microsoft/playwright-vscode/blob/959031d8e6ae5764ec8cdb088bf671800ab5e8b6/src/playwrightTest.ts#L47

mxschmitt:
Hi, thanks for trying out the extension. :) Could you send me the output of the debug logs? You can get them as shown in the screenshot:

![image](https://user-images.githubusercontent.com/17984549/130084083-1f25a269-6b49-4272-b7f2-8df63b19d6fa.png)

The path does not get passed over to the test-runner as a regular expression, it should be more like an exact string match of the file path. Once I have the output, I will take a more detailed look.

erwin-faceit:
At the moment it's like this (obfuscated, but you get the idea):

```
c:\Development\Project\Tests.Playwright\node_modules\@playwright\test\lib\cli\cli.js test --reporter=list c:\Development\Project\Tests.Playwright\tests\01-login-credentials.spec.ts:20
```

If I change this to the string below, it actually works.

```
c:\Development\Project\Tests.Playwright\node_modules\@playwright\test\lib\cli\cli.js test --reporter=list c:\\Development\\Project\\Tests.Playwright\\tests\\01-login-credentials\.spec\.ts:20
```

mxschmitt:
Fixed in #22, thanks for your contribution!

PR #22, opened 2021-08-19
Regexp escape paths passed to playwright (#21)

erwin-faceit:
I might have complicated the regular expression replace. I've got this one from https://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex.


mxschmitt (approved):
LGTM, thank you for your contribution! Don't forget to sign the CLA with the above link, otherwise we can't merge.

PR #23, opened 2021-08-20
Don't warn when Playwright Test not installed

### LINKED ISSUE #16: Do not warn that Playwright Test could not be found
Issue: Currently it does warn on every project that it can't be found when Playwright Test is not installed
Solution: Silently log it somewhere but no warning toast

mskelton:
Fixes #16

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -48,7 +48,7 @@ export async function activate(context: vscode.ExtensionContext) {
   try {
     playwrightTest = await PlaywrightTest.create(workspaceFolder.uri.fsPath, configuration.get('playwright.cliPath')!, debugModeHandler);
   } catch (error) {
-    vscode.window.showWarningMessage(error.toString());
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -48,7 +48,7 @@ export async function activate(context: vscode.ExtensionContext) {
   try {
     playwrightTest = await PlaywrightTest.create(workspaceFolder.uri.fsPath, configuration.get('playwright.cliPath')!, debugModeHandler);
   } catch (error) {
-    vscode.window.showWarningMessage(error.toString());
+    console.warn(error.toString());
```
mxschmitt, line 51: We have a logger, I think its logger.debug then so it gets showed inside the output panel.

mxschmitt (approved):

src/extension.ts
mskelton: Updated

PR #24, opened 2021-08-23
chore: enhance debugging experience

### LINKED ISSUE #18: Add Playwright Inspector integration
VSCode has two types of tests:
a) normal tests
-> there we want the inspector if the user enables it
b) debug tests
-> there we don't want the inspector, DEBUG=pw:api should be enough
-> because the VSCode debugger is attached

mxschmitt:
Fixes #18

### DIFF
--- a/src/playwrightTest.ts
+++ b/src/playwrightTest.ts
@@ -60,7 +60,7 @@ export class PlaywrightTest {
 
   public async runTest(config: PlaywrightTestConfig, projectName: string, path: string, line: number): Promise<playwrightTestTypes.JSONReport> {
     const proc = await this._executePlaywrightTestCommand(config, projectName, [`${escapeRegExp(path)}:${line}`], {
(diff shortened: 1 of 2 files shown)


src/playwrightTest.ts
```diff
@@ -85,13 +85,19 @@ export class PlaywrightTest {
     return result;
   }
 
-  private _getEnv(): NodeJS.ProcessEnv {
-    if (this._debugMode.isEnabled())
+  private _getEnv(vscodeDebuggerEnabled: boolean): NodeJS.ProcessEnv {
+    if (!this._debugMode.isEnabled())
+      return process.env;
+    // we don't want to have two debuggers at the same time
+    if (vscodeDebuggerEnabled)
```
aslushnikov, line 92: nit: curly braces around return since it's a multiline block

aslushnikov (approved):

PR #38, opened 2021-08-25
fix: only add test files to the test explorer

### LINKED ISSUE #31: Don't add every open JavaScript file to the Test Explorer
Currently every JS/TS file which gets opened gets added to the Test Explorer. Some validation should be necessary to filter out non-test files.

Code pointer: `updateNodeForDocument` function.

mxschmitt:
Fixes #31

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -132,9 +132,12 @@ async function createTestController(context: vscode.ExtensionContext, workspaceF
     if (!['.ts', '.js', '.mjs'].some(extension => e.uri.fsPath.endsWith(extension))) {
       return;
     }
-
-    const { file, data } = getOrCreateFile(ctrl, workspaceFolder, e.uri, playwrightTest, config, projectName);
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -132,9 +132,12 @@ async function createTestController(context: vscode.ExtensionContext, workspaceF
     if (!['.ts', '.js', '.mjs'].some(extension => e.uri.fsPath.endsWith(extension))) {
       return;
     }
-
-    const { file, data } = getOrCreateFile(ctrl, workspaceFolder, e.uri, playwrightTest, config, projectName);
-    data.updateFromDisk(ctrl, file);
+    playwrightTest.listTests(config, projectName, e.uri!.fsPath).then(tests => {
```
aslushnikov, line 135: this probably requires a critical section to fight against re-enterability

aslushnikov (approved):

PR #44, opened 2021-08-30
fix: get PW-Test output data via a file instead stdout

### LINKED ISSUE #19: [Bug] Extension does not work if something gets logged inside playwright.config

mxschmitt:
fixes #19

### DIFF
--- a/src/playwrightTest.ts
+++ b/src/playwrightTest.ts
@@ -14,14 +14,15 @@
  * limitations under the License.
  */
 
+import * as fs from 'fs';
 import * as path from 'path';
-import type { SpawnOptionsWithoutStdio } from 'child_process';
+import * as os from 'os';
 import * as vscode from 'vscode';
 
 import * as playwrightTestTypes from './testTypes';
 import { logger } from './logger';
 import type { PlaywrightDebugMode } from './extension';
-import { fileExistsAsync, spawnAsync, escapeRegExp } from './utils';
+import { fileExistsAsync, spawnAsync, escapeRegExp, createGuid } from './utils';
 
 export const DEFAULT_CONFIG = Symbol('default config');
 export type PlaywrightTestConfig = string | typeof DEFAULT_CONFIG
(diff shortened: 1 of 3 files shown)


src/playwrightTest.ts
```diff
@@ -45,31 +46,31 @@ export class PlaywrightTest {
...
-    const proc = await this._executePlaywrightTestCommand(config, projectName, [`${escapeRegExp(path)}:${line}`]);
+  public async runTest(config: PlaywrightTestConfig, projectName: string, testPath: string, line: number): Promise<playwrightTestTypes.JSONReport> {
+    const jsonOutputPath = path.join(os.tmpdir(), createGuid());
+    await this._executePlaywrightTestCommand(jsonOutputPath, config, projectName, [`${escapeRegExp(testPath)}:${line}`]);
     try {
-      return JSON.parse(proc.stdout);
-    } catch (error) {
-      logger.debug('could not parse JSON', proc.stdout, proc.stderr);
-      throw error;
+      return JSON.parse(await fs.promises.readFile(jsonOutputPath, 'utf8'));
```
mxschmitt, line 67: hmm, I should move this maybe into _executePlaywrightTestCommand,

mxschmitt: nvm, we want exit code validation only for listTests, so not that great.

src/playwrightTest.ts
```diff
@@ -45,31 +46,31 @@ export class PlaywrightTest {
...
+    } finally {
+      await fs.promises.unlink(jsonOutputPath);
     }
   }
 
-  public async runTest(config: PlaywrightTestConfig, projectName: string, path: string, line: number): Promise<playwrightTestTypes.JSONReport> {
-    const proc = await this._executePlaywrightTestCommand(config, projectName, [`${escapeRegExp(path)}:${line}`]);
+  public async runTest(config: PlaywrightTestConfig, projectName: string, testPath: string, line: number): Promise<playwrightTestTypes.JSONReport> {
+    const jsonOutputPath = path.join(os.tmpdir(), createGuid());
```
aslushnikov, line 64: let's prefix this file with something related to the name of this project:

```js
const jsonOutputPath = path.join(os.tmpdir(), 'playwright-vscode-test-exetension-' + createGuid());
```

this way it'll be possible to trace back some leftovers if any.

aslushnikov (approved):

Issue #50, opened 2021-09-02
Automatically enable the extension


mxschmitt:
if a user installs @playwright/test or if the users adds a playwright.config.ts

mxschmitt:
Folding into https://github.com/microsoft/playwright-vscode/issues/85

PR #53, opened 2021-09-08
chore: auto activate on playwright.config.{ts,js}

kumaraditya303:
Relates https://github.com/microsoft/playwright-vscode/issues/50

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
 		"Testing"
 	],
 	"activationEvents": [
+		"workspaceContains:playwright.config.{ts,js}",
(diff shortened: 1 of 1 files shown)


mxschmitt:
I think it does not fix it, but it does relate to it. Not all users have a Playwright config and its possible to use it without. So an activation based on if the dependency is installed would be the best.

kumaraditya303:
> I think it does not fix it, but it does relate to it. Not all users have a Playwright config and its possible to use it without. So an activation based on if the dependency is installed would be the best.

That can be added but playwright.config.ts is the option currently which confirms playwright test runner usage so this should be added.

PR #115, opened 2022-02-23
feat(message): render error message w/ colors

pavelfeldman:


src/utils.ts
```diff
@@ -74,9 +74,65 @@ export async function findInPath(program: string): Promise<string | undefined> {
...
+          break;
+        }
+        case '[32': {
+          tokens.push(`<span style='color:#f14c4c;'>`);
+          isOpen = true;
+          break;
+        }
+        case '[39': break;
+      }
+      i = end;
```
dgozman, line 111: `end` could be `-1` (`console.log('\u011bHELLO')`) and this will enter an infinite loop.

src/utils.ts
```diff
@@ -74,9 +74,65 @@ export async function findInPath(program: string): Promise<string | undefined> {
...
+        else
+          tokens.push(' ');
+      } else {
+        tokens.push(escapeHTML(c));
+        isLeadingSpace = false;
+      }
+    }
+  }
+  // Work around html parsing bugs.
+  if (hasTags)
```
dgozman, line 129: Why unpaired closing tag? Shouldn't this condition be `isOpen` instead?

pavelfeldman: As per comment above!

PR #134, opened 2022-03-02
feat(source-map): add support for compiled test run

pavelfeldman:


src/testModel.ts
```diff
@@ -168,11 +186,12 @@ export class TestModel {
   }
 
   async listTests(files: string[]) {
-    const filesToLoad = files.filter(f => this.allFiles.has(f));
-    if (!filesToLoad.length)
+    const sourcestoLoad = files.filter(f => this.allFiles.has(f));
```
mxschmitt, line 189:
```suggestion
    const sourcesToLoad = files.filter(f => this.allFiles.has(f));
```

src/utils.ts
```diff
@@ -150,3 +151,37 @@ export async function spawnAsync(executable: string, args: string[], cwd?: strin
     childProcess.on('exit', () => f(output));
   });
 }
+
+export async function resolveSourceMap(file: string, fileToSources: Map<string, string[]>, sourceToFile: Map<string, string>): Promise<string[]> {
+  if (!file.endsWith('.js'))
```
mxschmitt, line 156: mjs or cjs? probably not something where rollup or tsc compile to I guess?

src/utils.ts
```diff
@@ -150,3 +151,37 @@ export async function spawnAsync(executable: string, args: string[], cwd?: strin
...
+
+  const rl = readline.createInterface({ input: fs.createReadStream(file), crlfDelay: Infinity });
+
+  let lastLine: string | undefined;
+  rl.on('line', line => {
+    lastLine = line;
+  });
+  await new Promise(f => rl.on('close', f));
+
+  if (lastLine?.startsWith('//# sourceMappingURL=')) {
```
mxschmitt, line 170: what about inline sourcemaps? Not sure if they make sense here.

mxschmitt (approved):

src/utils.ts
pavelfeldman: We might need to support them at some point, but they are not the tsc/esbuild defaults.

Issue #144, opened 2022-03-18
[Bug] Improve handling of Playwright config throwing errors


weyert:
If you have a Playwright configuration that does some sanity checks, such as throwing an error when an environment variable is missing then the extension under the hood throws an error similar as listed below. I think it woud be nice to have a dedicated Playwright item in the `Output`-panel so you can easily spot issues when  executing the tests.

Maybe it could show such errors in a `showMessage`-dialog?

```
[2022-03-18 11:30:32.486] [renderer2] [error] [Extension Host] SyntaxError: Unexpected token i in JSON at position 0
    at JSON.parse (<anonymous>)
    at qe.listFiles (/Users/weyert/.vscode/extensions/ms-playwright.playwright-0.2.2/out/extension.js:14:3481)
[2022-03-18 11:30:34.407] [renderer2] [error] ENOENT: no such file or directory, stat 'node_modules': Error: ENOENT: no such file or directory, stat 'node_modules'
```

weyert:
I will see if I can solve this problem myself. I might ask some questions via Slack. Not well versed in VSCode Extension development (yet) :)

pavelfeldman:
@weyert: so you global setup throws. When global setup throws, reporter's `onError` method is called. So we should add a `onError` handler in `_runTest` that does the same thing `onStdErr` does (writes it into the testing console).

weyert:
While throwing Error's is definitely causing havoc and breaks things but when I remove that. 
It's still not working for me, I think the main cause of my problems is that I am using `console.log` in my project file which is causing trouble. As it will be part of the stdout when the extension is trying to query the available projects of the project.

The extension is expecting a JSON output for the spawned project for listing projects; that why I might a change in my PR to try to detect JSON blob and extract it from the output

pavelfeldman:
I finally understood what you were talking about! It is always easier to explain things in a form of test, especially given that you anyways will need one to go with the fix:

```js
test('should list files when config logs to stdout', async ({}, testInfo) => {
  const { testController, renderExecLog } = await activate(testInfo.outputDir, {
    'playwright.config.js': `
      console.log('bad JSON');
    `,
    'test.spec.ts': `
      import { test } from '@playwright/test';
      test('one', async () => {});
    `,
  });
  expect(testController.renderTestTree()).toBe(`
    - test.spec.ts
  `);
  expect(renderExecLog('  ')).toBe(`
    > playwright list-files -c playwright.config.js
  `);
});
```

PR #145, opened 2022-03-19
feat: allow the use of logging and process.cwd() in config files

weyert:
Improves the support of using `process.cwd()` and the `console.log` in
config files. Now all the spawned process of the extension will have the
working directory set to the workspace folder.

Next, also allow the use of `console.log` within the code that gets
executed in the context of the Playwright config file. Now it will try
to detect and parse the Project listing JSON blob out of the process stdout
and dismiss the other loggging

refs #144

### DIFF
--- a/src/playwrightTest.ts
+++ b/src/playwrightTest.ts
@@ -97,14 +97,21 @@ export class PlaywrightTest {
   async listFiles(config: TestConfig): Promise<ConfigListFilesReport | null> {
     const configFolder = path.dirname(config.configFile);
     const configFile = path.basename(config.configFile);
-    const allArgs = [config.cli, 'list-files', '-c', configFile];
+    const allArgs = [config.cli, 'list-files', '-c', config.configFile];
     {
       // For tests.
       this._log(`${escapeRegex(path.relative(config.workspaceFolder, configFolder))}> playwright list-files -c ${configFile}`);
     }
-    const output = await this._runNode(allArgs, configFolder);
+
+    // Execute the list-files command with workspace folder as working directory and
+    // use full path to the config file to allow process.cwd() to work as expected
+    // in config files
+    const output = await this._runNode(allArgs, config.workspaceFolder);
     try {
-      return JSON.parse(output) as ConfigListFilesReport;
+      // Strip the JSON blob starting with {"projects":} from the output of the
+      // executed process
+      const strippedOutput = /{"projects":(.+)\}/ig.exec(output)?.[0] ?? ''
+      return JSON.parse(strippedOutput) as ConfigListFilesReport;
     } catch (e) {
       console.error(e);
       return null;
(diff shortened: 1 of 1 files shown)


src/playwrightTest.ts
```diff
@@ -152,7 +159,7 @@ export class PlaywrightTest {
     if (mode === 'list')
       allArgs.push('--reporter', 'null');
     const childProcess = spawn(node, allArgs, {
-      cwd: configFolder,
+      cwd: config.workspaceFolder,
```
weyert, line 162: What was the reason to set the `cwd` for the process to the config directory?

src/playwrightTest.ts
```diff
@@ -97,14 +97,21 @@ export class PlaywrightTest {
...
+
+    // Execute the list-files command with workspace folder as working directory and
+    // use full path to the config file to allow process.cwd() to work as expected
+    // in config files
+    const output = await this._runNode(allArgs, config.workspaceFolder);
     try {
-      return JSON.parse(output) as ConfigListFilesReport;
+      // Strip the JSON blob starting with {"projects":} from the output of the
+      // executed process
+      const strippedOutput = /{"projects":(.+)\}/ig.exec(output)?.[0] ?? ''
```
weyert, line 113: Really hacky but it works in my circumstance. I am wondering if there is a better way to exchange the list of projects

src/playwrightTest.ts
```diff
@@ -203,7 +210,7 @@ export class PlaywrightTest {
       type: 'pwa-node',
       name: debugSessionName,
       request: 'launch',
-      cwd: configFolder,
+      cwd: config.workspaceFolder,
       env: {
         ...process.env,
```
weyert, line 215: I am wondering if for all these processes that get spawned if it would make sense to also set a _special_ environment variable something similar to `IS_PLAYWRIGHT=1` so people could change behaviours when code is running inside Playwright context

pavelfeldman: We could, but I am not sure we want to encourage that behavior. Do you have a good example of the conditional behavior?

src/playwrightTest.ts
pavelfeldman: What exactly are we fixing here? Let's get issues filed for those two problems?

src/playwrightTest.ts
weyert: Good point, not sure, originally I was thinking I could stop throwing error's in this situation but that wouldn't be a good way to solve things. I will try to add your suggested `onError`-listener

src/playwrightTest.ts
```diff
@@ -97,14 +97,21 @@ export class PlaywrightTest {
...
     {
       // For tests.
       this._log(`${escapeRegex(path.relative(config.workspaceFolder, configFolder))}> playwright list-files -c ${configFile}`);
     }
-    const output = await this._runNode(allArgs, configFolder);
+
+    // Execute the list-files command with workspace folder as working directory and
+    // use full path to the config file to allow process.cwd() to work as expected
+    // in config files
+    const output = await this._runNode(allArgs, config.workspaceFolder);
```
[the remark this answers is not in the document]
weyert: Hmm, good point, probably I should drill down to the first `package.json` in my project and use that path instead of depending on `process.cwd()`. Didn't think of that 🤔

weyert:
> What exactly are we fixing here? Let's get issues filed for those two problems?

We are fixing the issue described here: https://github.com/microsoft/playwright-vscode/issues/144#issuecomment-1076941015

src/playwrightTest.ts
pavelfeldman: We resolve everything relative to that directory, for example when nested folders resolve to different versions of Playwright.

pavelfeldman:
Closing as stale

PR #171, opened 2022-06-08
fix: handle whitespace is error message

JonahKK:
When the whitespace is the difference between Expected and Received value, the values are not properly displayed in the error message:
Image 08-06-2022 at 08 22

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -100,12 +100,12 @@ export function ansiToHtml(text: string): string {
         }
         case '[22': break;
         case '[31': {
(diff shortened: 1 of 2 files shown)


pavelfeldman:
Did you run it to see if it helps? My understanding is that this fix won't only leave the problem as is, it'll also reset the remaining color highlighting. VS Code sanitizes the styles and adding white-space makes it think the the style is bad.

pavelfeldman:
This was addressed using & nbsp ;

PR #185, opened 2022-07-04
chore: have better defaults during install

mxschmitt:
We should always enable GitHub Actions by default, and install the linux dependencies as well. Both either breaks the user story/flow or makes the browsers not launchable.


mxschmitt:
offline review with pfeldman

PR #191, opened 2022-07-06
chore: use in process 'which' instead of spawning sub process

mxschmitt:
It seems popular, maintained and better than spawning a subprocess see [here](https://www.npmjs.com/package/which).

This fixes the flaky tests too.


mxschmitt:
approved via TG by pf.

Issue #97, opened 2022-02-17
two refresh buttons when playwright and python extensions are enabled


DetachHead:
![image](https://user-images.githubusercontent.com/57028336/154376563-dfdb37d8-af27-4b84-b3b1-a599b7350821.png)

pavelfeldman:
Does the python button work for you?

connor4312:
In the linked issue, there is a now [standard API for this](https://github.com/microsoft/vscode/issues/139737) which will be stable in our next release (<= 14 days from today)

weyert:
Yeah, I am also seeing to Refresh buttons. If I press on it I am getting an error talking about pytests

KotlinIsland:
Talking about getting an error message, I need to see the full-form of the message or I can't fix it!

PR #192, opened 2022-07-06
chore: unite refresh button with the one from testing API

mxschmitt:
Fixes https://github.com/microsoft/playwright-vscode/issues/97

Was added in VSCode 1.65: https://code.visualstudio.com/updates/v1_65#_testing-refresh-action-and-sorttext

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -99,6 +99,14 @@ export class Extension {
     this._recorder = new Recorder(this._vscode, this._playwrightTest);
     this._testController = vscode.tests.createTestController('pw.extension.testController', 'Playwright');
     this._testController.resolveHandler = item => this._resolveChildren(item);
+    this._testController.refreshHandler = () => {
+      this._rebuildModel(true).then(configs => {
+        if (!configs.length) {
+          vscode.window.showWarningMessage('No Playwright Test config files found.');
+          return;
+        }
+      }).catch();
+    };
     this._testTree = new TestTree(vscode, this._testController);
     this._debugHighlight = new DebugHighlight(vscode);
     this._debugHighlight.onErrorInDebugger(e => this._errorInDebugger(e.error, e.location));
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -47,19 +47,18 @@ type TestRunInfo = {
 
 export async function activate(context: vscodeTypes.ExtensionContext) {
   // Do not await, quickly run the extension, schedule work.
-  new Extension(require('vscode')).activate(context);
+  new Extension(require('vscode')).activate(context).catch();
```
pavelfeldman, line 50: Why the `.catch()`?

src/extension.ts
```diff
@@ -97,14 +96,33 @@ export class Extension {
...
     this._workspaceObserver = new WorkspaceObserver(this._vscode, changes => this._workspaceChanged(changes));
   }
 
+  private ensureTestTree(): TestTree {
+    if (this._testTree)
+      return this._testTree;
+    const testController = this._vscode.tests.createTestController('pw.extension.testController', 'Playwright');
+    testController.resolveHandler = item => this._resolveChildren(item);
+    testController.refreshHandler = () => {
+      this._rebuildModel(true).catch();
```
pavelfeldman, line 110: why the `.catch()`?

src/extension.ts
```diff
@@ -97,14 +96,33 @@ export class Extension {
 
     this._playwrightTest = new PlaywrightTest(this._isUnderTest);
     this._recorder = new Recorder(this._vscode, this._playwrightTest);
-    this._testController = vscode.tests.createTestController('pw.extension.testController', 'Playwright');
-    this._testController.resolveHandler = item => this._resolveChildren(item);
-    this._testTree = new TestTree(vscode, this._testController);
     this._debugHighlight = new DebugHighlight(vscode);
```
pavelfeldman, line 99: I don't think we need these (debug highlight + workspace observer) before we get our first config.

I wonder if a better approach would be to install the `**/*playwright*.config.{ts,js,mjs}` watcher and activate the entire extension based on whether we have configs. I.e. keep the extension object as is, but wrap it into something that'll activate and dispose it.

PR #200, opened 2022-08-03
feat(server): introduce browser server mode

pavelfeldman:


src/playwrightTest.ts
```diff
@@ -222,6 +229,52 @@ export class PlaywrightTest {
...
+        return;
+      this._browserServerWS = match[1];
+    });
+
+    stdio[2].on('data', data => {
+      console.log('Error in browser server: ' + data.toString());
+    });
+
+    token.onCancellationRequested(() => {
+      serverProcess.stdin.write('<EOL>');
```
dgozman, line 263: I think we used to `stdin.close()` in other places instead? That would just imitate disconnect.

src/playwrightTest.ts
```diff
@@ -241,6 +294,17 @@ export class PlaywrightTest {
     return node;
   }
 
+  async findNpx(): Promise<string> {
+    if (this._pathToNpxJS)
+      return this._pathToNpxJS;
+
+    const npx = await which('npx');
+    if (!npx)
+      throw new Error('Unable to launch `node`, make sure it is in your PATH');
```
dgozman, line 303:
```suggestion
      throw new Error('Unable to launch `npx`, make sure it is in your PATH');
```

src/playwrightTest.ts
```diff
@@ -222,6 +229,52 @@ export class PlaywrightTest {
     await reporterServer.wireTestListener(listener, token);
   }
 
+  async runBrowserServer(workspaceFolder: string, token: vscodeTypes.CancellationToken) {
+    if (this._browserServerWS)
+      return;
+
+    const npx = await this.findNpx();
+
+    const allArgs = [
```
dgozman, line 238: If you do `node playwright-core/cli.js ...`, that will save a process and a second.

src/playwrightTest.ts
```diff
@@ -149,17 +151,22 @@ export class PlaywrightTest {
       '--repeat-each', '1',
       '--retries', '0',
     ];
-    if (this._isUnderTest)
+    if (this._isUnderTest || !!this._browserServerWS)
```
dgozman, line 154: Should we check that playwright versions from `config` and `runBrowserServer` match and error otherwise?

dgozman (approved):

src/playwrightTest.ts
pavelfeldman: Unfortunately, it does not. I think we were doing it in our Java port only.

src/playwrightTest.ts
pavelfeldman: That would mean that we have a multi-workspace configuration in VS Code or a monorepo where different projects have different versions of Playwright. I'd say ignore this edge case for now? I should add a minimum version check for run-server though, I'll do that.

src/playwrightTest.ts
pavelfeldman: I originally was trying to implement it in a model-less manner, but I guess it is ok.

PR #208, opened 2022-08-09
chore: remove headed setting

pavelfeldman:


mxschmitt (approved):

src/reusedBrowser.ts
```diff
@@ -169,6 +179,8 @@ test('test', async ({ page }) => {
   }
 
   async willRunTests(config: TestConfig) {
+    if (!this._checkVersion(config, 'Show & reuse browser'))
```
dgozman, line 182: Why is this here, and not in `startIfNeeded`? Do we only get here if "show & reuse" is checked?

dgozman (approved):

PR #213, opened 2022-08-11
chore: cancel previous mode on new mode

pavelfeldman:


src/reusedBrowser.ts
```diff
@@ -188,27 +196,29 @@ test('test', async ({ page }) => {
...
-  private async _cleanupModes() {
+  private async _reset() {
     // This won't wait for setMode(none).
-    if (this._selectorExplorerBox) {
-      this._selectorExplorerBox.dispose();
-      this._selectorExplorerBox = undefined;
-    }
+    this._updateOrCancelInspecting?.({ cancel: true });
+    this._updateOrCancelInspecting = undefined;
+    this._cancelRecording?.();
```
dgozman, line 211: This will call `_reset()` again in line 166, but I guess it's fine?

dgozman (approved):

src/reusedBrowser.ts
pavelfeldman: Yeah, should be microtasked...

PR #218, opened 2022-08-15
feat(update): suggest installing browsers on update

pavelfeldman:


src/installer.ts
```diff
@@ -86,3 +87,62 @@ export async function installPlaywright(vscode: vscodeTypes.VSCode) {
...
+    description: '— powers Mozilla Firefox',
+  };
+  const webkit: vscodeTypes.QuickPickItem = {
+    label: 'WebKit',
+    picked: true,
+    description: '— powers  Apple Safari',
+  };
+  const installDepsAction: vscodeTypes.QuickPickItem = {
+    label: 'Install Linux dependencies',
+    picked: process.platform === 'linux' && !fs.existsSync(path.join(process.env.XDG_CACHE_HOME || path.join(os.homedir(), '.cache'), 'ms-playwright')),
```
yury-s, line 109: why do we need this `!fs.existsSync(path.join(process.env.XDG_CACHE_HOME` ?

src/extension.ts
```diff
@@ -582,7 +594,9 @@ located next to Run / Debug Tests toolbar buttons.`);
   }
 
   private _testMessageForTestError(testItem: vscodeTypes.TestItem, error: TestError): vscodeTypes.TestMessage {
-    const text = error.stack || error.message || error.value!;
+    let text = error.stack || error.message || error.value!;
+    if (text.includes('Looks like Playwright Test or Playwright'))
+      text = `Browser was not installed. Invoke 'Install Playwright Browsers' action to install missing browsers.`;
```
yury-s, line 599: will `Install Playwright Browsers` actually match the action ?

yury-s (approved):

src/installer.ts
pavelfeldman: yes!

src/extension.ts
pavelfeldman: Yes

src/installer.ts
```diff
@@ -73,16 +52,87 @@ export async function installPlaywright(vscode: vscodeTypes.VSCode) {
...
+    args.push('chromium');
+  if (result.includes(firefoxItem))
+    args.push('firefox');
+  if (result.includes(webkitItem))
+    args.push('webkit');
+
+  if (args.length)
+    terminal.sendText(`npx playwright ${installCommand} ${args.join(' ')}`, true);
+  else if (result.includes(installDepsItem))
+    terminal.sendText(`npx playwright install-deps`, true);
```
mxschmitt, line 108: Sudo?

pavelfeldman: We don't do that currently.

PR #219, opened 2022-08-16
feat(close): allow closing all the browsers

pavelfeldman:


src/extension.ts
```diff
@@ -148,6 +148,9 @@ export class Extension {
         }
         await this._reusedBrowser.inspect(this._models);
       }),
+      vscode.commands.registerCommand('pw.extension.command.closeBrowsers', () => {
+        this._reusedBrowser.stop();
```
rwoll, line 152: Does VSCode allow you to hook into close lifecycle of the VSCode itself and trigger this there too? I got in state where I had closed VSCode and then I had to close browser after manually.

rwoll (approved):
Assuming it's been tested locally, LGTM.

src/extension.ts
pavelfeldman: That would be very strange - I thought this part of the pipeline is supersolid. When VSCode goes, driver self-desctructs, taking all the browsers with itself.

PR #228, opened 2022-08-24
fix: do not scan node_modules

kalinkrustev:
There is no reason to scan node_modules in the workspace and then ignore files. This also allows the extension to work when your workspace folder path includes node_modules.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -198,7 +198,7 @@ export class Extension {
     this._workspaceObserver.reset();
     this._models = [];
 
-    const configFiles = await this._vscode.workspace.findFiles('**/*playwright*.config.{ts,js,mjs}', 'node_modules/**');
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -198,16 +198,14 @@ export class Extension {
     this._workspaceObserver.reset();
     this._models = [];
 
-    const configFiles = await this._vscode.workspace.findFiles('**/*playwright*.config.{ts,js,mjs}');
+    const configFiles = await this._vscode.workspace.findFiles('**/*playwright*.config.{ts,js,mjs}', 'node_modules/**');
```
mxschmitt, line 201:
```suggestion
    const configFiles = await this._vscode.workspace.findFiles('**/*playwright*.config.{ts,js,mjs}', '**/node_modules/**');
```

richmar10:
I know this is closed, but I'm pretty sure we're also checking the .git folder by accident.

PR #229, opened 2022-08-24
chore: detect pnp workspaces and lookup playwright-core using yarn's PnP API

steschi:
Since there have been some issues regarding repositories which do not offer a traditional `node_modules` folder (#133)  I tried to get this extension working for repositories with yarn-v3 and PnP. 

Did not find a contribution guide so let me know if something needs to change or if you do not accept any foreign contributions.

### DIFF
--- a/src/playwrightTest.ts
+++ b/src/playwrightTest.ts
@@ -21,7 +21,7 @@ import { Entry, StepBeginParams, StepEndParams, TestBeginParams, TestEndParams }
 import type { TestError } from './reporter';
 import { ReporterServer } from './reporterServer';
 import { ReusedBrowser } from './reusedBrowser';
-import { findNode, spawnAsync } from './utils';
+import { findNode, getPnPEnvVariables, isPnPWorkspace, spawnAsync } from './utils';
 import * as vscodeTypes from './vscodeTypes';
 
 export type TestConfig = {
@@ -66,6 +66,30 @@ export class PlaywrightTest {
   }
 
   async getPlaywrightInfo(workspaceFolder: string, configFilePath: string): Promise<{ version: number, cli: string } | null> {
+
+    if (isPnPWorkspace(workspaceFolder)) {
+      try {
+        const pwtCoreInfo = await this._runNode([
+          '-e',
+          'const pnpapi = require("pnpapi"); ' +
+          'const locators = pnpapi.getAllLocators().filter(locator => locator.name === "playwright-core"); ' +
+          'if (locators.length === 0) { process.exit(1); } ' +
+          'const info = pnpapi.getPackageInformation(locators[0]); ' +
+          'console.log(JSON.stringify({version: locators[0].reference, ...info}));'
+        ],
+        path.dirname(configFilePath),
+        getPnPEnvVariables(workspaceFolder));
+
+        const { packageLocation, version } = JSON.parse(pwtCoreInfo);
+        return {
+          cli: path.join(packageLocation, 'lib/cli/cli'),
+          version: parseFloat(version.replace('npm:', ''))
+        };
+      } catch (error) {
+        console.log('unable to use pnp.cjs', error);
+      }
+    }
+
     try {
       const pwtInfo = await this._runNode([
         '-e',
@@ -104,7 +128,7 @@ export class PlaywrightTest {
       // For tests.
       this._log(`${escapeRegex(path.relative(config.workspaceFolder, configFolder))}> playwright list-files -c ${configFile}`);
     }
-    const output = await this._runNode(allArgs, configFolder);
+    const output = await this._runNode(allArgs, configFolder, getPnPEnvVariables(config.workspaceFolder));
     try {
       return JSON.parse(output) as ConfigListFilesReport;
     } catch (e) {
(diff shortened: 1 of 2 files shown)


src/playwrightTest.ts
```diff
@@ -66,6 +66,30 @@ export class PlaywrightTest {
   }
 
   async getPlaywrightInfo(workspaceFolder: string, configFilePath: string): Promise<{ version: number, cli: string } | null> {
+
+    if (isPnPWorkspace(workspaceFolder)) {
+      try {
+        const pwtCoreInfo = await this._runNode([
+          '-e',
+          'const pnpapi = require("pnpapi"); ' +
```
pavelfeldman, line 74: Is this the package you are using? https://www.npmjs.com/package/pnpapi

steschi: No, in yarn-v3 PnP environments `pnpapi` is a built-in module. The package on npm is just a placeholder and not used.  
Details can be found here [https://yarnpkg.com/advanced/pnpapi#requirepnpapi](https://yarnpkg.com/advanced/pnpapi#requirepnpapi)

steschi:
Is this just a CI hiccup or can I do something to fix the tests on windows node16? Locally the tests all pass (I am on Windows with node16).

src/utils.ts
```diff
@@ -88,11 +88,32 @@ function escapeHTML(text: string): string {
...
+export function isPnPWorkspace(workspaceDir: string): boolean {
+  return fs.existsSync(getPnPFilepath(workspaceDir));
+}
+export function getPnPEnvVariables(workspaceDir: string): Record<string,string> {
+  if (!isPnPWorkspace(workspaceDir))
+    return {};
+
+  const pnpFile = getPnPFilepath(workspaceDir);
+  return {
+    NODE_OPTIONS: `--require ${pnpFile}`
```
mxschmitt, line 103:
```suggestion
    NODE_OPTIONS: `${process.env.NODE_OPTIONS ? (process.env.NODE_OPTIONS + ' ') : ''}--require ${pnpFile}`
```

steschi: Do you know if there could be other sources of `NODE_OPTIONS` other than `process.env`?
in `src/playwrightTests.ts` `debugTests`:
```js
  ...process.env,
  ...settingsEnv,
  ...this._reusedBrowser.browserServerEnv(),
  ...(await reporterServer.env()),
```
Can any of these provide additional `NODE_OPTIONS`? I think your suggestion could lead to options being applied twice

mxschmitt: NODE_OPTIONS can come from a variety of different places, `process.env` is very likely to have some, `settingsEnv` can also contain some. While the other two most likely not.

I think it's fine to merge all 3 into a single one, Node.js would deal with that in the end and prioritizes the right ones.

src/playwrightTest.ts
```diff
@@ -66,6 +66,30 @@ export class PlaywrightTest {
...
+          'const locators = pnpapi.getAllLocators().filter(locator => locator.name === "playwright-core"); ' +
+          'if (locators.length === 0) { process.exit(1); } ' +
+          'const info = pnpapi.getPackageInformation(locators[0]); ' +
+          'console.log(JSON.stringify({version: locators[0].reference, ...info}));'
```
mxschmitt, line 78:
```suggestion
          `const pnpapi = require('pnpapi');
          const pwCoreLocator = pnpapi.getAllLocators().find(locator => locator.name === 'playwright-core');
          if (!pwCoreLocator)
            process.exit(1);
          const info = pnpapi.getPackageInformation(pwCoreLocator);
          console.log(JSON.stringify({version: pwCoreLocator.reference, ...info}));`
```

that should work as well I guess?

src/utils.ts
```diff
@@ -88,11 +88,32 @@ function escapeHTML(text: string): string {
...
+  const env = {
+    ...process.env,
+    ...additionalEnv
+  };
+
   const childProcess = spawn(executable, args, {
     stdio: 'pipe',
     cwd,
-    env: { ...process.env }
+    env
```
mxschmitt, line 116:
```suggestion
    env: {
      ...process.env,
      ...additionalEnv,
   }
```

looks simpler enough that we can keep it inline.

src/playwrightTest.ts
```diff
@@ -66,6 +66,30 @@ export class PlaywrightTest {
   }
 
   async getPlaywrightInfo(workspaceFolder: string, configFilePath: string): Promise<{ version: number, cli: string } | null> {
+
+    if (isPnPWorkspace(workspaceFolder)) {
```
mxschmitt, line 70: would be also good if we can move it out to a different file, add some more handlers, hooks etc. so playwrightTest.ts is not so much involved.

src/utils.ts
steschi: I am just a little hesitant about merging all these as, in my understanding, only the last `NODE_OPTIONS` would survive?

src/playwrightTest.ts
steschi: So you would like to move all pnp-related utility functions into your own file?

What do you mean by "why node -e does not work"?

PR #239, opened 2022-09-16
chore: support docker integration

aslushnikov:


mxschmitt:
GC, we'll revisit at some point.

PR #251, opened 2022-10-13
Allow user-settings to override NODE_OPTIONS

badeball:
Change in accordance with https://github.com/microsoft/playwright/issues/17925 (@pavelfeldman).

Fixes https://github.com/microsoft/playwright/issues/17925

### DIFF
--- a/src/playwrightTest.ts
+++ b/src/playwrightTest.ts
@@ -166,12 +166,12 @@ export class PlaywrightTest {
       stdio: ['pipe', 'pipe', 'pipe', 'pipe', 'pipe'],
       env: {
         ...process.env,
-        ...this._reusedBrowser.browserServerEnv(false),
-        ...this._reusedBrowser.logApiCallsEnv(),
-        ...(await reporterServer.env()),
         // Don't debug tests when running them.
         NODE_OPTIONS: undefined,
         ...settingsEnv,
+        ...this._reusedBrowser.browserServerEnv(false),
+        ...this._reusedBrowser.logApiCallsEnv(),
+        ...(await reporterServer.env()),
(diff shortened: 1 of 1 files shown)


src/playwrightTest.ts
```diff
@@ -166,12 +166,12 @@ export class PlaywrightTest {
       stdio: ['pipe', 'pipe', 'pipe', 'pipe', 'pipe'],
       env: {
         ...process.env,
-        ...settingsEnv,
         ...this._reusedBrowser.browserServerEnv(false),
         ...this._reusedBrowser.logApiCallsEnv(),
         ...(await reporterServer.env()),
         // Don't debug tests when running them.
         NODE_OPTIONS: undefined,
+        ...settingsEnv,
```
pavelfeldman, line 174: Let's move NODE_OPTIONS to be before ...settingsEnv instead. That way users are not accidentally override the other internal env variables we rely upon (lines 169-171 new code).

SanariSan: Thanks for mentioning `NODE_OPTIONS`.
Because of that, I was able to find this pull and trace it back to `this._vscode.workspace.getConfiguration('playwright').get('env', {});`, which led to figuring out how to pass those `NODE_OPTIONS` from vscode.

P.S. This is what I did. I needed that because of the decorators issue mentioned in #29646.
`.vscode/settings.json`
```json
{
    "playwright.env": {
        "NODE_ENV": "test",
        "TS_NODE_PROJECT": "./test/tsconfig.json",
        "NODE_OPTIONS": "--require ts-node/register/transpile-only --require tsconfig-paths/register",
    }
}
```

PR #264, opened 2022-11-01
chore: make record from here record from cursor

pavelfeldman:


src/reusedBrowser.ts
```diff
@@ -160,17 +165,38 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
...
+            this._insertedEditActionCount = params.actions.length;
+            if (params.actions?.length) {
+              await this._editor.edit(async editBuilder => {
+                if (!this._editor)
+                  return;
+                const indentation = guessIndentation(this._editor);
+                const action = params.actions[params.actions.length - 1].trim();
+                const lineNumber = this._editor.selection.start.line;
+                const selection = new this._vscode.Selection(new this._vscode.Position(lineNumber, 0), this._editor.selection.end);
+                editBuilder.replace(selection, indentation + action + '\n');
```
dgozman, line 186: How does this work with multiline actions?

pavelfeldman: It does not, in a sense that they'll have poor formatting. I can follow up when I am able to add tests.

PR #268, opened 2022-11-03
feat(highlight): highlight when not debugging as well

pavelfeldman:


src/methodNames.ts
```diff
@@ -71,5 +71,9 @@ export const locatorMethods = [
   'getByTestId',
   'getByText',
   'getByTitle',
-  'getBy'
+  'getBy',
```
dgozman, line 74: There is no `getBy` :)

dgozman (approved):

PR #277, opened 2022-11-22
add support for yarn pnp

orkisz:
This is an attempt to resolve the issue described here: https://github.com/microsoft/playwright/issues/18931

I am aware of yet another PR that tries to solve the problem (https://github.com/microsoft/playwright-vscode/pull/229), but I took a slightly different approach here.

All functions I've tested works, I have problem with some tests failing.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -206,7 +206,7 @@ export class Extension {
     this._workspaceObserver.reset();
     this._models = [];
 
-    const configFiles = await this._vscode.workspace.findFiles('**/*playwright*.config.{ts,js,mjs}', '**/node_modules/**');
+    const configFiles = await this._vscode.workspace.findFiles('**/*playwright*.config.{ts,js,mjs}', '**/{node_modules,.yarn}/**');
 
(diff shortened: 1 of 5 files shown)


pavelfeldman:
I closed the original issue because supporting non-Node runtimes is outside of the scope of Playwright at this point.

chasestarr:
Hi, I see that the https://github.com/microsoft/playwright/issues/18931 issue has been closed but I'm curious why this and https://github.com/microsoft/playwright-vscode/pull/229 PRs are still open. Is the team simply not interested in supporting yarn pnp or are there reservations to the particular PRs? I [made a fork with support](https://github.com/chasestarr/playwright-vscode-yarn-pnp/tree/main) for myself, but I'm happy to contribute here if the team would prefer an alternate approach to solving the issue

mxschmitt:
We landed https://github.com/microsoft/playwright-vscode/pull/301 for now as a minimal implementation and will follow-up with auto detection in the future.

See https://github.com/microsoft/playwright-vscode/pull/301 for how to use it once we released a new version.

PR #312, opened 2023-02-25
Fix Install Playwright Browser with deps command

fflaten:
`Install Playwright Browser` command currently fails when choosing to include linux dependencies.

```
$ npx playwright install-with-deps chromium firefox webkit
error: unknown command 'install-with-deps'
```

PR updated the command to call use `install --with-deps` which is the correct command for (at least) 1.19+


mxschmitt (approved):
Thanks!

PR #314, opened 2023-03-03
fix(output): wait 500ms before starting the test run

dgozman:
This allows us to create our own test request, and also produce test output.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -300,7 +300,7 @@ export class Extension {
     usedProfiles.add(debugProfile);
   }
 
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -320,11 +320,14 @@ export class Extension {
     if (!this._projectsScheduledToRun) {
       this._projectsScheduledToRun = [];
       this._projectsScheduledToRun.push(project);
-      await Promise.resolve().then(async () => {
-        const selectedProjects = this._projectsScheduledToRun!;
+      // Make sure to run tests outside of this function's control flow
```
pavelfeldman, line 323: Say TODO remove in 1.78?

pavelfeldman (approved):

PR #322, opened 2023-04-04
support l10n with vscode-l10n

csbun:
[UPDATED at Apr 20, 2023]
- Translations are placed in `l10n/bundle.l10n.{LANG}.json` folder. Currently, `zh-CN` is supported.
- Use [vscode-l10n](https://github.com/microsoft/vscode-l10n) for l10n. 
- Run `npm run l10n` will search all TypeScript files in `./src` and place a `l10n/bundle.l10n.json` file.

[Mar 30, 2023]
- ~~Translations are placed in `i18n/${language}` folder. Currently, `zh-CN (chs)` is supported.~~
- ~~Use [vscode-nls](https://github.com/Microsoft/vscode-nls) for i18n.~~ 
- ~~Build with gulp and [vscode-nls-dev](https://github.com/Microsoft/vscode-nls-dev). Running `npm run build` will generate `package.nls.*.json` and `out/nls.*.json`.~~


csbun:
@microsoft-github-policy-service agree

mxschmitt (commented):
Sorry for the late review. Would be nice if we could have the 

```
nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
const localize: nls.LocalizeFunc = nls.loadMessageBundle();
```

a single time.

Also would be cool if you could revert the vscode.d.ts changes. Maybe we can get rid of gulp? Will then review further once this is done.

csbun:
1. Maybe we can create a file like `src/localize.ts` including those codes above and exporting `localize` as default. Then we can import it from any other files.
2. I had added a dependence `@types/vscode` in package.json [L109](https://github.com/microsoft/playwright-vscode/pull/322/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R109) so we may need to keep the vscode.d.ts file. Or is there any reason we had to maintain it ourselves?
3. [vscode-nls](https://github.com/Microsoft/vscode-nls) and vscode-nls-dev [had just left a deprecated message a few days ago](https://github.com/microsoft/vscode-nls/commit/6f88c23c3ab630e9d4c60f65ed33839bd4a0cec2), which suggest using glup officially. I will try to rewrite it with [vscode-l10n](https://github.com/microsoft/vscode-l10n) later.

tsconfig.json
```diff
@@ -8,8 +8,7 @@
     "strict": true,
     "rootDir": "src",
     "esModuleInterop": true,
-    "noEmit": true
```
mxschmitt, line 11: was this change intended?

tsconfig.json
```diff
@@ -8,8 +8,7 @@
...
   },
-  "include": ["src", "vscode.d.ts"],
+  "include": ["src"],
   "exclude": ["node_modules"]
```
mxschmitt, line 13: would be cool if we could keep the vscode.d.ts for now inside this PR like we had before. We are happy to get rid of it I believe, but if so, then in a separate PR. Thanks!

package.json
```diff
@@ -107,11 +109,13 @@
...
     "@types/node": "^18.11.9",
     "@types/stack-utils": "^2.0.1",
+    "@types/vscode": "^1.73.0",
     "@types/which": "^2.0.1",
     "@types/ws": "^8.5.3",
     "@typescript-eslint/eslint-plugin": "^5.44.0",
     "@typescript-eslint/parser": "^5.44.0",
-    "esbuild": "^0.15.15",
+    "@vscode/l10n-dev": "^0.0.24",
+    "esbuild": "^0.17.17",
```
mxschmitt, line 118: I believe the esbuild version bump was not related.

csbun: sure. will rollback to ^0.15.15

tsconfig.json
csbun: Thanks for pointing out the mistake.

tsconfig.json
csbun: Another PR sounds more reasonable.

package.nls.json
```diff
@@ -0,0 +1,13 @@
+{
+  "description": "Run Playwright Test tests in Visual Studio Code. from nls",
```
mxschmitt, line 2:
```suggestion
  "description": "Run Playwright Test tests in Visual Studio Code.",
```

package.nls.json
```diff
@@ -0,0 +1,13 @@
...
+  "contributes.command.pw.extension.install": "Install Playwright",
+  "contributes.command.pw.extension.installBrowsers": "Install Playwright Browsers",
+  "contributes.command.pw.extension.command.inspect": "Pick locator",
+  "contributes.command.pw.extension.command.closeBrowsers": "Close all Playwright browsers",
+  "contributes.command.pw.extension.command.recordNew": "Record new",
+  "contributes.command.pw.extension.command.recordAtCursor": "Record at cursor",
+  "contributes.command.pw.extension.toggle.reuseBrowser": "Toggle Playwright Browser Reuse",
+  "configuration.playwright.env": "Environment variables to pass to Playwright Test.",
+  "configuration.playwright.reuseBrowser": "Show & reuse browser between tests.",
+  "views.test.pw.extension.settingsView": "Playwright"
```
mxschmitt, line 12: this (settingsView) needs to get used inside the package.json as well.

src/extension.ts
```diff
@@ -135,6 +136,7 @@ export class Extension {
   async activate(context: vscodeTypes.ExtensionContext) {
     const vscode = this._vscode;
     this._settingsView = new SettingsView(vscode, this._settingsModel, this._reusedBrowser, context.extensionUri);
+    const messageNoPlaywrightFound = l10n.t('No Playwright tests found.');
```
mxschmitt, line 139:
```suggestion
    const messageNoPlaywrightTestsFound = l10n.t('No Playwright tests found.');
```

src/extension.ts
```diff
@@ -235,14 +237,20 @@ export class Extension {
...
+        }
         continue;
       }
 
       if (playwrightInfo.version < 1.28) {
-        if (showWarnings)
-          this._vscode.window.showWarningMessage('Playwright Test v1.28 or newer is required');
+        if (showWarnings) {
+          this._vscode.window.showWarningMessage(
+              l10n.t('Playwright Test v1.28 or newer is required')
```
mxschmitt, line 251: since we are already touching this code, lets add a variable `const minimumPlaywrightVersion = 1.28` in line 248, use it inside the if and pass it over to the translation message as a variable. Thanks!

mxschmitt (commented):
good progress!

mxschmitt:
Thank you!

PR #330, opened 2023-05-12
chore: use @types/vscode dependence instead of a vscode.d.ts file

csbun:


csbun:
@mxschmitt any update on this PR?

PR #335, opened 2023-05-25
Open UI Mode command

tonybaloney:
Request from Debbie O'Brien.

Added a command to start UI mode from VS Code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,11 @@
       },
       {
         "category": "Test",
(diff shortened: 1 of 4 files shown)


mxschmitt:
Thank you for your contribution! We implemented "Live Trace" since a couple of releases in order to solve this. Having UI Mode integrated in VSCode we felt is not the best idea UX wise, since you then have two test tree views.

PR #339, opened 2023-05-25
fix: allow settings action clicks next to label

mxschmitt:
The issue is that customers often click on the text next to the label, which has `cursor: pointer`, so they think it will open codegen, but in reality, this is not doing anything, because there is no click handler attached.

image

This is the issue which a bunch of our lab users encountered during the demo at MSBuild.

### DIFF
--- a/media/settingsView.css
+++ b/media/settingsView.css
@@ -67,6 +67,7 @@ label {
   display: flex;
   align-items: center;
   cursor: inherit;
+  flex: auto;
 }
 
 div.separator {


src/settingsView.ts
```diff
@@ -169,8 +169,8 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
               actionElement.setAttribute('disabled', 'true');
             const label = document.createElement('label');
             if (!action.disabled) {
-              label.addEventListener('click', event => {
-                vscode.postMessage({ method: 'execute', params: { command: event.target.getAttribute('command') } });
+              actionElement.addEventListener('click', event => {
```
pavelfeldman, line 172: You could as well just flex: auto the label?

pavelfeldman (approved):

PR #344, opened 2023-06-08
chore: friendlier uikind.web operation

pavelfeldman:


src/reusedBrowser.ts
```diff
@@ -333,6 +333,14 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
       );
       return false;
     }
+
+    if (this._vscode.env.uiKind === this._vscode.UIKind.Web) {
```
mxschmitt, line 337: Maybe instead when the user toggles the button? Or grey it out instead?

mxschmitt (approved):

src/reusedBrowser.ts
pavelfeldman: Hiding it is a bit more involved.

dgozman (approved):

PR #362, opened 2023-07-18
fix: show browser always in non-web mode

mxschmitt:
Partial revert of https://github.com/microsoft/playwright-vscode/commit/d6881ef874f94f499a2be8f70e2d978e0babc847

For future reference:

telegram-cloud-photo-size-2-5312353643384657502-y

We keep the remoteName check in our trace viewer decision logic.

Fixes https://github.com/microsoft/playwright/issues/24254

### DIFF
--- a/src/reusedBrowser.ts
+++ b/src/reusedBrowser.ts
@@ -334,7 +334,7 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
       return false;
     }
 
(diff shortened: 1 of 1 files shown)


pavelfeldman (approved):

yugawara:
Thank you you for this revert - so show Browser should be Yes for container (and all other than codespaces)?

mxschmitt:
This should work again after this PR yes.

PR #366, opened 2023-07-28
chore: do not show can not close message

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/24491

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
     "@types/glob": "^8.0.0",
     "@types/node": "^18.11.9",
     "@types/stack-utils": "^2.0.1",
-    "@types/vscode": "1.73.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -118,7 +118,7 @@
     "@types/glob": "^8.0.0",
     "@types/node": "^18.11.9",
     "@types/stack-utils": "^2.0.1",
-    "@types/vscode": "1.73.0",
+    "@types/vscode": "1.80.0",
```
mxschmitt, line 121: Related?

mxschmitt (approved):

package.json
pavelfeldman: It is ok to have it!

PR #378, opened 2023-09-15
fix: best-effort support for POMs when highlighting

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/27036

Adding a test was more involved than I thought. We already have some logic in place to connect to an existing "ReusableBrowser" but the issue is that `waitForPage` calls `_newContextForReuse` which will reset it, and I want to keep the old content. https://github.com/microsoft/playwright-vscode/blob/ba3258a4a23b0391596fe21164b46c2ad9061d8e/tests/utils.ts#L71-L87

I went a different route over CDP instead.

### DIFF
--- a/src/debugHighlight.ts
+++ b/src/debugHighlight.ts
@@ -147,6 +147,8 @@ async function locatorToHighlight(document: vscodeTypes.TextDocument, position: 
     });
     // Translate locator expressions starting with "component." to be starting with "page.".
     locatorExpression = locatorExpression?.replace(/^component\s*\./, `page.locator('#root').locator('internal:control=component').`);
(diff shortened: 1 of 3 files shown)


src/debugHighlight.ts
```diff
@@ -23,9 +23,8 @@ import * as vscodeTypes from './vscodeTypes';
 
 export type DebuggerError = { error: string, location: Location };
 
-const debugSessions = new Map<string, vscodeTypes.DebugSession>();
-
 export class DebugHighlight {
+  private _debugSessions = new Map<string, vscodeTypes.DebugSession>();
```
mxschmitt, line 27: note: reason for moving it out was that the tests were not isolated before, the `_debugSession` struct leaked.

dgozman (approved):

PR #398, opened 2023-12-01
fix: double fills while debugging

mxschmitt:
This regressed in https://github.com/microsoft/playwright-vscode/commit/59f828d54ffa6e8a06620bdb894009d35b500279.

Fixes https://github.com/microsoft/playwright/issues/28437.

### DIFF
--- a/src/methodNames.ts
+++ b/src/methodNames.ts
@@ -73,12 +73,10 @@ export const locatorMethods = [
   'getByTitle',
   'first',
   'last',
+  'and',
+  'or',
   'nth',
   'filter',
-  'check',
-  'click',
-  'fill',
-  'type',
 ];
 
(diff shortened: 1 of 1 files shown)


mxschmitt:
> Can we keep the highlight?

I tested highlight of `await page.fill('[placeholder="What needs to be done?"]', TODO_ITEMS[0]);` in step mode (breakpoint) and highlight mode (move mouse over it). Seems to work in both cases.

pavelfeldman (approved):

nbaldzhiev:
Hi @mxschmitt @pavelfeldman , how can I track what version this will be included in? I have that bug as well in VS Code Version: 1.84.2 (Universal) Commit: 1a5daa3a0231a0fbba4f14db7ec463cf99d7768e on MacOS 13.4.1 (22F82), Playwright is at `1.37.1` and I am currently debugging in Chromium.

PR #399, opened 2023-12-27
French localization

jfgreffier:
Translations for French

Language code is simply `fr`, as I believe it is standard French and there are no needs for region-specific language (belgium, swiss, french canada, ...)


mxschmitt (approved):

mxschmitt:
Thanks!

PR #400, opened 2024-01-09
fix: evaluate 'which' via shell if Node.js is not in PATH (fallback)

mxschmitt:
Relates https://github.com/microsoft/playwright/issues/28703

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -138,18 +138,47 @@ export async function resolveSourceMap(file: string, fileToSources: Map<string, 
 
 let pathToNodeJS: string | undefined;
 
-export async function findNode(): Promise<string> {
+export async function findNode(vscode: typeof import('vscode')): Promise<string> {
   if (pathToNodeJS)
     return pathToNodeJS;
 
+  // Stage 1: try to find node via process.env.PATH
   let node = await which('node').catch(e => undefined);
-  // When extension host boots, it does not have the right env set, so we might need to wait.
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -138,18 +138,47 @@ export async function resolveSourceMap(file: string, fileToSources: Map<string,
...
+  // Stage 3: If we still haven't found node, try to find it via a subprocess.
+  // This evaluates shell rc/profile files and makes nvm work.
+  node ??= await findNodeViaSubProcess(vscode);
   if (!node)
     throw new Error('Unable to launch `node`, make sure it is in your PATH');
   pathToNodeJS = node;
   return node;
 }
+
+async function findNodeViaSubProcess(vscode: typeof import('vscode')): Promise<string | undefined> {
```
mxschmitt, line 161:
```suggestion
async function findNodeViaShell(vscode: typeof import('vscode')): Promise<string | undefined> {
```

nit?

src/utils.ts
```diff
@@ -138,18 +138,47 @@ export async function resolveSourceMap(file: string, fileToSources: Map<string,
 
 let pathToNodeJS: string | undefined;
 
-export async function findNode(): Promise<string> {
+export async function findNode(vscode: typeof import('vscode')): Promise<string> {
```
pavelfeldman, line 141: Here and below,

`vscode: vscode.VSCode`

pavelfeldman (approved):

PR #402, opened 2024-01-10
chore: show more debug info when no Node.js is available

mxschmitt:
https://github.com/microsoft/playwright/issues/28703

Looks like that now in DevTools AND gets surfaced in VSCode as a warning.

Screenshot 2024-01-10 at 11 00 40

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -26,7 +26,7 @@ import { SettingsModel } from './settingsModel';
 import { SettingsView } from './settingsView';
 import { TestModel, TestProject } from './testModel';
 import { TestTree } from './testTree';
-import { ansiToHtml } from './utils';
+import { NodeJSNotFoundError, ansiToHtml } from './utils';
 import * as vscodeTypes from './vscodeTypes';
 import { WorkspaceChange, WorkspaceObserver } from './workspaceObserver';
 import { TraceViewer } from './traceViewer';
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -202,7 +202,7 @@ export class Extension {
       this._diagnostics,
       this._treeItemObserver,
     ];
-    await this._rebuildModel(false);
```
yury-s, line 205: Why did this change?

yury-s (approved):

src/extension.ts
mxschmitt: This function gets called when the extension gets activated. For the initial activation we want to show the error messages, what this boolean is about: `showWarnings: true`.

src/extension.ts
```diff
@@ -240,13 +240,16 @@ export class Extension {
...
+      try {
+        playwrightInfo = await this._playwrightTest.getPlaywrightInfo(workspaceFolderPath, configFilePath);
+      } catch (error) {
         if (showWarnings) {
           this._vscode.window.showWarningMessage(
-              this._vscode.l10n.t('Please install Playwright Test via running `npm i --save-dev @playwright/test`')
+              error instanceof NodeJSNotFoundError ? error.message : this._vscode.l10n.t('Please install Playwright Test via running `npm i --save-dev @playwright/test`')
           );
         }
+        console.error('[Playwright Test]:', (error as any)?.message, 1);
```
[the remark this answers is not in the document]
mxschmitt: I like to keep this error which we consider fatal in DevTools, like we had it before. If the user did e.g. disable all the warnings from the extension, what he can, this would not be shown and hidden everywhere.

PR #406, opened 2024-01-15
Fix findNodeViaShell to run within the workspace folder

slushie:
This change sets the `cwd` option for `spawn()` to be the current workspace folder when spawning a shell in `findNodeViaShell`. Without this change, the search will run from within the current directory of the parent process, which is `/` by default. This PR makes it possible for project-local node version configurations to be used with the Playwright plugin.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -167,9 +167,11 @@ async function findNodeViaShell(vscode: vscodeTypes.VSCode): Promise<string | un
   return new Promise<string | undefined>(resolve => {
     const startToken = '___START_PW_SHELL__';
     const endToken = '___END_PW_SHELL__';
+    const cwd = vscode.workspace.workspaceFolders ? vscode.workspace.workspaceFolders[0] : undefined;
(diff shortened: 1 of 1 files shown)


mxschmitt:
Looks reasonable, just to make sure. Was this a regression for you? Since this method is new and was implemented as a fallback, probably not?

mxschmitt:
> > Looks reasonable, just to make sure. Was this a regression for you? Since this method is new and was implemented as a fallback, probably not?
> 
> > 
> 
> > I created #407 so its always based on the workspace folder of the current test project.
> 
> 
> 
> Nice, thank you! <3
> 
> 
> 
> I am a new Playwright user, so I don't know offhand if this is a regression with the recent related changes. But with my setup based on per-directory config, I think its most likely that this wouldn't have worked at all before. 
> 
> 
> 
> In fact, I was looking at making a follow-up PR (adding `runtimeExecutable: findNode(...)` [near here](https://github.com/microsoft/playwright-vscode/blob/main/src/playwrightTest.ts#L228-L250)) to handle the debug launcher which suffers the same problem of not finding the `node` executable. I'm happy to contribute that as well, but I think you've got the gist of it here ;)

Thank you! Sounds like a great plan!

PR #413, opened 2024-02-05
fix: only mark first project as default

mxschmitt:
https://github.com/microsoft/playwright/issues/29367

As per https://code.visualstudio.com/updates/v1_86#_finalized-testrunprofileisdefaultondidchangedefault-apis-for-extension-authors.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -234,7 +234,6 @@ export class Extension {
 
     const configFiles = await this._vscode.workspace.findFiles('**/*playwright*.config.{ts,js,mjs}', '**/node_modules/**');
 
-    let isFirstProject = true;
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -281,16 +279,10 @@ export class Extension {
       }
 
       for (const project of model.projects.values()) {
-        await this._createRunProfile(project, usedProfiles);
+        await this._createRunProfile(project, isFirstProject);
```
pavelfeldman, line 282: Now every time we rebuild the model user settings will be reset. We should restore them.

pavelfeldman: They also used to persist across the sessions, I wonder what happened to that use case.

PR #419, opened 2024-02-15
chore: extract backend server util

pavelfeldman:


src/backend.ts
```diff
@@ -0,0 +1,135 @@
...
+        ...process.env,
+        ...this._envProvider(),
+      },
+    });
+
+    const client = this._clientFactory();
+
+    serverProcess.stderr?.on('data', () => {});
+    serverProcess.on('error', error => {
+      client?._onErrorEvent.fire(error);
```
dgozman, line 53:
```suggestion
      client._onErrorEvent.fire(error);
```

src/backend.ts
```diff
@@ -0,0 +1,135 @@
...
+        const match = data.toString().match(/Listening on (.*)/);
+        if (!match)
+          return;
+        const wse = match[1];
+        await client._connect(wse);
+        fulfill(client);
+      });
+      serverProcess.on('exit', () => {
+        fulfill(null);
+        client?._onCloseEvent.fire();
```
dgozman, line 67:
```suggestion
        client._onCloseEvent.fire();
```

src/backend.ts
```diff
@@ -0,0 +1,135 @@
...
+        error.stack = message.error.error?.stack;
+        pair.reject(error);
+      } else {
+        pair.fulfill(message.result);
+      }
+    };
+    await this.initialize();
+  }
+
+  async initialize() { }
```
dgozman, line 123: Perhaps we can do `abstract` here to force subclasses to implement it.

dgozman (approved):

src/backend.ts
pavelfeldman: not everybody needs it.

PR #421, opened 2024-02-21
chore: optionally use test server

pavelfeldman:


src/testServerController.ts
```diff
@@ -0,0 +1,72 @@
...
+      dumpIO: false,
+    });
+    const testServer = await testServerBackend.start();
+    if (!testServer)
+      return null;
+    this._testServers.set(config, testServer);
+    return testServer;
+  }
+
+  dispose() {
```
dgozman, line 50: Perhaps I missed, but seems like noone is calling this.

dgozman (approved):

PR #423, opened 2024-02-22
chore: use test server as a singleton

pavelfeldman:


src/testServerController.ts
```diff
@@ -47,9 +53,7 @@ export class TestServerController implements vscodeTypes.Disposable {
       dumpIO: false,
     });
     const testServer = await testServerBackend.start();
-    if (!testServer)
-      return null;
-    this._testServers.set(config, testServer);
+    this._instance = testServer;
```
dgozman, line 56: Seems to be unused.

dgozman (approved):

PR #437, opened 2024-03-13
chore: select projects in sidebar

pavelfeldman:


src/testModel.ts
```diff
@@ -258,9 +320,145 @@ export class TestModel {
...
+      for (const project of model.projects()) {
+        const projectSettings = configSettings.projects.find(p => p.name === project.name);
+        if (projectSettings)
+          project.isEnabled = projectSettings.enabled;
+        else if (firstProject)
+          project.isEnabled = true;
+        firstProject = false;
+      }
+    } else {
+      model.projects()[0].isEnabled = true;
```
dgozman, line 399: Should we find first non-dependency project, like in UI mode?

dgozman (approved):

src/testModel.ts
pavelfeldman: Good call. I did not add code / tests for dependencies yet.

PR #442, opened 2024-03-15
feat: allow filtering for native VSCode Testing API tags

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/29938
Relates https://github.com/microsoft/vscode/issues/207877

### DIFF
--- a/src/testTree.ts
+++ b/src/testTree.ts
@@ -144,6 +144,8 @@ export class TestTree extends DisposableBase {
         vsChildrenById.set(id, vsChild);
         vsChildren.add(vsChild);
       }
+      if (uChild.kind === 'case')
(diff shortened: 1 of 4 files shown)


src/testTree.ts
```diff
@@ -144,6 +144,8 @@ export class TestTree extends DisposableBase {
         vsChildrenById.set(id, vsChild);
         vsChildren.add(vsChild);
       }
+      if (uChild.kind === 'case')
+        vsChild.tags = uChild.tags.map(tag => new this._vscode.TestTag(tag));
```
pavelfeldman, line 148: Let's compare them to the existing ones first in order to not trigger update every time.

pavelfeldman (approved):

PR #445, opened 2024-03-22
chore: use uimode's test server api

pavelfeldman:


src/testModel.ts
```diff
@@ -57,7 +58,7 @@ export class TestModel {
   private _envProvider: () => NodeJS.ProcessEnv;
   isEnabled = false;
   readonly tag: vscodeTypes.TestTag;
-  private _configError: reporterTypes.TestError | undefined;
+  private _errors = new MultiMap<string, reporterTypes.TestError>();
```
dgozman, line 61: To make it obvious what the key is:

```suggestion
  private _errorsByFile = new MultiMap<string, reporterTypes.TestError>();
```

src/testModel.ts
```diff
@@ -225,25 +224,33 @@ export class TestModel {
...
+      if (error.location)
+        this._errors.set(error.location.file, error);
+    }
+
     for (const [projectName, project] of this._projects) {
       const files = projectFiles(project);
       const newProjectSuite = newProjectSuites.find(e => e.project()!.name === projectName);
       const filesToClear = new Set(requestedFiles);
       for (const fileSuite of newProjectSuite?.suites || []) {
+        if (this._errors.has(fileSuite.location!.file))
```
dgozman, line 248: I wonder how can we get a non-empty suite for a file that has a loading error?

src/testServerController.ts
```diff
@@ -59,65 +58,43 @@ export class TestServerController implements vscodeTypes.Disposable {
...
+      onError: error => {
+        this._connectionPromises.delete(config.configFile);
+      },
     });
-    const testServer = await testServerBackend.start();
+    if (!wsEndpoint)
+      return null;
+    const testServer = new TestServerConnection(wsEndpoint);
+    await testServer.connect();
+    this._connectionPromises.set(config.configFile, Promise.resolve(testServer));
```
dgozman, line 73: Do you want to set the promise at the start of this function, to avoid races when calling it twice before connection is done?

src/backend.ts
```diff
@@ -143,3 +119,34 @@ export class BackendClient extends EventEmitter {
...
+    serverProcess.stdout?.on('data', async data => {
+      if (options.dumpIO)
+        console.log('[server out]', data.toString());
+      const match = data.toString().match(/Listening on (.*)/);
+      if (!match)
+        return;
+      const wse = match[1];
+      fulfill(wse);
+    });
+    serverProcess.on('exit', () => {
```
dgozman, line 148: It looks like `options.onClose` is not wired.

dgozman (approved):

src/testModel.ts
pavelfeldman: that would be a subsequent load upon watch

src/testServerController.ts
pavelfeldman: Oh yes

PR #448, opened 2024-03-27
chore: only list tests when needed

pavelfeldman:


src/testModel.ts
```diff
@@ -225,32 +236,48 @@ export class TestModel {
...
-          allFilesLoaded = false;
-          break;
-        }
-      }
-      if (!allFilesLoaded)
-        break;
+    if (changed.length) {
+      for (const c of changed)
+        this._filesWithListedTests.delete(c);
+      await this.ensureTests(changed);
```
dgozman, line 242: This makes us list tests for all changed files, e.g. on `git checkout`. Perhaps we should only re-list in those files that are already in the `_filesWithListedTests`?

tests/project-setup.spec.ts
```diff
@@ -121,3 +121,22 @@ test('should run part of the setup only', async ({ activate }) => {
         ◯ test [2:0]
   `);
 });
+
+test('should run setup and teardown for test', async ({ activate }) => {
+  test.fixme();
```
dgozman, line 126: Ouch! Sounds like a regression compared to before TestServer?

dgozman (approved):

src/testModel.ts
pavelfeldman: Let me do that.

PR #452, opened 2024-04-01
chore: update watches to the new model structure

pavelfeldman:


src/extension.ts
```diff
@@ -311,11 +310,18 @@ export class Extension implements RunHooks {
...
       return;
     }
 
-    await this._runTests(request.include, isDebug ? 'debug' : 'run');
+    await this._queueTestRun(request.include, isDebug ? 'debug' : 'run');
+  }
+
+  private async _queueTestRun(include: readonly vscodeTypes.TestItem[] | undefined, mode: 'run' | 'debug' | 'watch') {
+    this._runQueue = this._runQueue.then(async () => {
+      await this._runTests(include, mode);
```
dgozman, line 322: `try/catch` just in case it fails and breaks the promise chain?

dgozman (approved):

PR #454, opened 2024-04-02
chore: add existing connection log to tests

pavelfeldman:


tests/mock/vscode.ts
```diff
@@ -846,12 +844,13 @@ export class VSCode {
...
   private _hoverProviders: Map<string, HoverProvider> = new Map();
   readonly version: string;
+  readonly connectionLog: any[] = [];
 
   constructor(readonly versionNumber: number, baseDir: string, browser: Browser) {
     this.version = String(versionNumber);
     this.context = { subscriptions: [], extensionUri: Uri.file(baseDir) };
     this._browser = browser;
-
+    (globalThis as any).__logForTest = message => this.connectionLog.push(message);
```
dgozman, line 853: Is this an upstream feature? I cannot find a second usage of `__logForTest`.

dgozman (approved):

tests/mock/vscode.ts
pavelfeldman: it is upstream yes

PR #464, opened 2024-04-25
chore: remove internal options, enable watch

pavelfeldman:


src/testModel.ts
```diff
@@ -698,10 +698,17 @@ export class TestModelCollection extends DisposableBase {
...
+    if (!enabledModels.length) {
+      this._selectedConfigFile = undefined;
+      return undefined;
+    }
+
+    const model = enabledModels.find(m => m.config.configFile === this._selectedConfigFile);
     if (model)
       return model;
-    return this._models.find(m => m.isEnabled);
+    this._selectedConfigFile = enabledModels[0].config.configFile;
```
dgozman, line 710: Updating in a getter sounds strange! Perhaps the method that enables/disables a model should update this field? Or default to a first enabled model when `this._selectedConfigFile` points nowhere?

dgozman (approved):

src/testModel.ts
pavelfeldman: > Or default to a first enabled model when this._selectedConfigFile points nowhere?

That's what it was, but in that case picking new model will change the current selection in the UI, which does not sound friendly. Think about this assignment here as a cache.

PR #483, opened 2024-05-18
feature(trace-viewer): embedded mode support PoC

ruifigueira:
PoC of trace-viewer embedded in VSCode.

It allows to open trace zip files as well as live traces of running tests.

https://github.com/microsoft/playwright-vscode/assets/1374559/e21cb7d6-8406-45ee-9de8-431b240e0ebb

Depends on https://github.com/microsoft/playwright/pull/30885


pavelfeldman:
I'm supportive for those, let me know when they are no longer drafts!

ruifigueira:
It now supports toggle between dark / light theme:


https://github.com/microsoft/playwright-vscode/assets/1374559/fa6ec224-04a3-4944-9487-d13f99dc8c6a

ruifigueira:
I added a few tests. They require a playwright version with changes from https://github.com/microsoft/playwright/pull/30885 , and that's why I left them as **wip** for now.

ruifigueira:
It's now working with vscode SSH remote dev and dev containers.

One nice thing is that there's no need to expose playwright server as `0.0.0.0` because trace viewer is opened by vscode and therefore we can take advantage of [vscode.env.asExternalUri](https://code.visualstudio.com/api/advanced-topics/remote-extensions#option-1-use-asexternaluri) to perform a localhost port forwarding.

ruifigueira:
It also works in codespaces:

https://github.com/microsoft/playwright-vscode/assets/1374559/13480576-4908-4de8-961a-ccef90198ca5

src/extension.ts
```diff
@@ -253,6 +259,18 @@ export class Extension implements RunHooks {
     this._context.subscriptions.push(this);
   }
 
+  private async _pickTraceFile() {
+    const vscode = this._vscode;
+    const selectedUris = await vscode.window.showOpenDialog({
+      canSelectMany: false,
+      filters: {
+        'Traces': ['zip'],
```
pavelfeldman, line 267: Time for us to claim .pwtrace upstream, so that we could just add a VS Code viewer.

src/extension.ts
```diff
@@ -748,20 +766,37 @@ export class Extension implements RunHooks {
...
+  private async _showTrace(testItem: vscodeTypes.TestItem) {
     const traceUrl = (testItem as any)[traceUrlSymbol];
     const testModel = this._models.selectedModel();
     if (testModel)
-      this._traceViewer.open(traceUrl, testModel.config);
+      this._traceViewer.open(traceUrl, await this._extendTestConfig(testModel));
   }
 
-  private _treeItemSelected(treeItem: vscodeTypes.TreeItem | null) {
+  private async _treeItemSelected(treeItem: vscodeTypes.TreeItem | null) {
```
pavelfeldman, line 776: Making event handlers async is not nice since we end up with unhandled rejections. Let's try to keep things sync.

src/extension.ts
```diff
@@ -748,20 +766,37 @@ export class Extension implements RunHooks {
...
     if (!treeItem)
       return;
     const traceUrl = (treeItem as any)[traceUrlSymbol] || '';
     const testModel = this._models.selectedModel();
     if (testModel)
-      this._traceViewer.open(traceUrl, testModel.config);
+      this._traceViewer.open(traceUrl, await this._extendTestConfig(testModel));
+  }
+
+  private async _fileSelected(uri?: vscodeTypes.Uri) {
```
pavelfeldman, line 785: ditto

src/extension.ts
```diff
@@ -535,7 +553,7 @@ export class Extension implements RunHooks {
     if (isDebug) {
       await model.debugTests(items, testListener, testRun.token);
     } else {
-      await this._traceViewer.willRunTests(model.config);
+      await this._traceViewer.willRunTests(await this._extendTestConfig(model));
```
pavelfeldman, line 556: It seems like moving the trace viewer instantiation into PlaywrightTestCLI / PlaywrightTestServer would be a meaningful refactoring before this change. Something like

```
if (this._traceViewer)
  this._traceViewer.close();
this._traceViewer = await model.showTraceViewerIfNeeded(); // delegates to PlaywrightTest{CLI/Server}
await model.runTests(items, testListener, testRun.token);
```

Models will run in-order sequentially, so before each model run (that corresponds to a specific test sever), you'll be opening trace viewer using that exact test server.

src/settingsModel.ts
```diff
@@ -62,6 +64,10 @@ export class SettingsModel extends DisposableBase {
       if (enabled && this.showBrowser.get())
         this.showBrowser.set(false);
     });
+    this.embedTraceViewer.onChange(enabled => {
+      if (enabled && !this.showTrace.get())
+        this.showTrace.set(true);
```
pavelfeldman, line 69: Let's keep it orthogonal to Show trace for now, as a separate setting.

src/settingsView.ts
```diff
@@ -312,6 +320,19 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
           vscode.postMessage({ method: 'toggle', params: { setting: event.target.getAttribute('setting') } });
         });
       }
+
+      function updateEmbedTraceViewer() {
```
pavelfeldman, line 324: Any reason this is not a part of updateSettings?

src/settingsView.ts
```diff
@@ -374,24 +398,20 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
...
+              updateEmbedTraceViewer();
             }
           }
           select.addEventListener('change', event => {
             vscode.postMessage({ method: 'selectModel', params: { configFile: select.value } });
             updateProjects(configsMap.get(select.value).projects);
           });
           const modelSelector = document.getElementById('model-selector');
-          modelSelector.style.display = showModelSelector ? 'block' : 'none';
+          if (showModelSelector)
```
pavelfeldman, line 409: modelSelector.classList.toggle('hidden', !showModelSelector);

src/testModel.ts
```diff
@@ -83,14 +83,18 @@ export class TestModel {
   constructor(vscode: vscodeTypes.VSCode, workspaceFolder: string, configFile: string, playwrightInfo: { cli: string, version: number }, options: TestModelOptions) {
     this._vscode = vscode;
     this._options = options;
-    this.config = { ...playwrightInfo, workspaceFolder, configFile };
     this._useLegacyCLIDriver = playwrightInfo.version < 1.44;
     this._playwrightTest =  this._useLegacyCLIDriver ? new PlaywrightTestCLI(vscode, this, options) : new PlaywrightTestServer(vscode, this, options);
     this._didUpdate = new vscode.EventEmitter();
     this.onUpdated = this._didUpdate.event;
+    this.config = { ...playwrightInfo, workspaceFolder, configFile };
```
pavelfeldman, line 90: Why did this move?

src/testModel.ts
```diff
@@ -83,14 +83,18 @@ export class TestModel {
...
     this.tag = new this._vscode.TestTag(this.config.configFile);
   }
 
+  async serverUrlPrefix() {
+    return this._playwrightTest instanceof PlaywrightTestServer ? await this._playwrightTest.serverUrlPrefix() : undefined;
```
pavelfeldman, line 95: Let's further encapsulate it as per comment above.

src/playwrightTestServer.ts
```diff
@@ -298,14 +307,14 @@ export class PlaywrightTestServer {
...
-    return this._testServerPromise;
+  private async _testServer() {
+    if (!this._testServerPromise)
+      this._testServerPromise = this._createTestServer();
+    const [server] = await this._testServerPromise;
+    return server;
   }
 
-  private async _createTestServer(): Promise<TestServerConnection | null> {
+  private async _createTestServer(): Promise<[TestServerConnection, string] | []> {
```
pavelfeldman, line 317: Don't do return tuples, keep the state in the PlaywrightTestServer instance, you should not need to expose it further.

src/playwrightTestServer.ts
```diff
@@ -365,11 +375,13 @@ export class PlaywrightTestServer {
...
 
-  private _disposeTestServer() {
+  private async _disposeTestServer() {
     const testServer = this._testServerPromise;
     this._testServerPromise = undefined;
-    if (testServer)
-      testServer.then(server => server?.close());
+    if (!testServer)
+      return;
+    const [server] = await testServer ?? [];
```
pavelfeldman, line 383: Let's keep it sync for clarity, vscode dispose is sync in nature.

tests/mock/vscode.ts
```diff
@@ -1077,6 +1100,88 @@ export class VSCode {
       d.dispose();
   }
 
+  async webViewsByPanelType(viewType: string) {
+    return (await Promise.all(this._webViewsByPanelType.get(viewType) ?? [])).filter(p => !p.isClosed());
+  }
+
+  private _createWebview(): Webview & { pagePromise?: Promise<Page>, show: () => Promise<Page>} {
```
pavelfeldman, line 1107: Let's extract / review / land the testing harness separately, before your change.

pavelfeldman (commented):
Looks like a good start, I reviewed everything except for actual trace viewer / web view and the tests.

src/testModel.ts
ruifigueira: I was initially tring to add `serverUrlPrefix` into the config and for that I needed the `PlaywrightTestServer`. But then I realized that I could only get the value asynchronously and aborted that approach, but I forgot to undo the move. I'll rollback it

src/extension.ts
ruifigueira: Ok, understood. That way, I can even pass the `model.config` to traceViewer constructor, instead of passing it in each call.

I was also considering splitting `TraceViewer` into two implementations: one with the legacy logic and another with the embedded logic. I think it simplifies the code (right now, I find it a little messy). And with your new proposal, it will be easy to ensure that we start the right TraceViewer implementation each time.

src/settingsView.ts
ruifigueira: That function hides the **Embed Trace Viewer** when the project is from an older version, but now that I look deeper into it, an alternative is to keep that option always visible but show a warning and fallback to the spawn process version if the project doesn't support embedded mode. That is in line with `showTrace` setting, that is always shown even when the project uses an old playwright version.

src/extension.ts
ruifigueira: I didn't understand what should I do there, can you clarify?

tests/mock/vscode.ts
ruifigueira: Should I create a new PR for it?

src/extension.ts
ruifigueira: 7644d39 contains your proposal and the trace viewer split. Due to the nature of current `TraceViewer` (I renamed it to `SpawnTraceViewer`), which is shared among testModels, its lyfecycle is still managed in `extension.ts`.

Each model can provide an optional `EmbededTraceViewer` via `PlaywrightTestServer`. Then when we need a trace viewer we check the settings and if the model supports embedded mode, otherwise we fallback to `SpawnTraceViewer`.

This way, we can keep the current use case of multiple test models using the same `SpawnTraceViewer` without needed to close and open it each time.

ruifigueira: I'm testing scenarios with multiple test configs. When running a test set from multiple test configs, it iterates over each test config, as you said, but it will close embedded trace viewer too many times and unnecessarily. That doesn't happen when we just select test items with traces because it always uses the selected test model in those cases.

I think we could do the same on `Extension._runTest`: instead of using each test config to pick the corresponding trace viewer, we always use the selected test config one.

tests/mock/vscode.ts
ruifigueira: I created a new PR: https://github.com/microsoft/playwright-vscode/pull/503

package.json
```diff
@@ -70,6 +70,11 @@
         "command": "pw.extension.toggle.showTrace",
         "title": "%contributes.command.pw.extension.toggle.showTrace%"
       },
+      {
+        "category": "Test",
+        "command": "pw.extension.toggle.embedTraceViewer",
```
pavelfeldman, line 75:
```suggestion
        "command": "pw.extension.toggle.embeddedTraceViewer",
```

src/extension.ts
```diff
@@ -208,6 +207,12 @@ export class Extension implements RunHooks {
       vscode.commands.registerCommand('pw.extension.command.clearCache', async () => {
         await this._models.selectedModel()?.clearCache();
       }),
+      vscode.commands.registerCommand('pw.extension.command.openTrace', async (uri?: vscodeTypes.Uri) => {
```
pavelfeldman, line 210: Let's add this separately.

src/extension.ts
pavelfeldman: That was not meant as a call for changes. I was saying that if we had a better extension, `foo.pwtrace`, we would be able to implement a VS Code viewer for it.

src/extension.ts
```diff
@@ -482,10 +500,8 @@ export class Extension implements RunHooks {
           return;
 
         const trace = result.attachments.find(a => a.name === 'trace')?.path || '';
-        // if trace viewer is currently displaying the trace file about to be replaced, it needs to be refreshed
-        const prevTrace = (testItem as any)[traceUrlSymbol];
         (testItem as any)[traceUrlSymbol] = trace;
-        if (enqueuedSingleTest || prevTrace === this._traceViewer.currentFile())
+        if (enqueuedSingleTest)
```
pavelfeldman, line 504: Can we land this refactoring separately?

src/playwrightTestCLI.ts
```diff
@@ -37,6 +39,22 @@ export class PlaywrightTestCLI {
...
 
   reset() {
+    if (!this._traceViewers)
+      return;
+    for (const traceViewer of this._traceViewers)
+      traceViewer.dispose();
+    this._traceViewers = undefined;
+  }
+
+  async availableTraceViewers(): Promise<TraceViewer[]> {
```
pavelfeldman, line 49: Why not traceViewer() that would return enabled instance? You don't seem to save much via having the enabled check on the call site. Also, if those are shallow objects, why not created them in the constructor? It would read better.

src/traceViewer.ts
```diff
@@ -17,67 +17,79 @@
...
-      await this._startIfNeeded(config);
+  isStarted() {
+    return !!this._traceViewerProcess;
   }
 
-  async open(file: string, config: TestConfig) {
-    if (!this._settingsModel.showTrace.get())
-      return;
-    if (!this._checkVersion(config))
+  async willRunTests() {
```
pavelfeldman, line 63: I'd try pulling this to the call site as well. Think about it this way: SpawnTraceViewer and EmbedTraceViewer know as little about the embedder as possible, including the settings. An external entity picks the viewer based on the setting, calls its willRunTests, etc.

src/traceViewer.ts
```diff
@@ -17,67 +17,79 @@
...
+  return typeof uriOrPath === 'string' ?
+    uriOrPath :
+    uriOrPath.scheme === 'file' ?
+      uriOrPath.fsPath :
+      uriOrPath.toString();
+}
+
+export type TraceViewer = SpawnTraceViewer | EmbeddedTraceViewer;
+
+export class SpawnTraceViewer extends DisposableBase {
```
pavelfeldman, line 34: Let's put these into separate files. When splitting the change, we can land SpawnTraceViewer in one PR and follow up with adding EmbeddedTraceViewer. That way we would see if there is something funny with the chosen modularity.

pavelfeldman (commented):
The change looks good. Please split it into the chunks so that I could do thorough review. These look like good chunks, but open to variations:

- WebView support (nonce)
- Selection strategy (selected item)
- Trace viewer abstraction + ShallowTraceViewer
- Additional embedded trace viewer

src/extension.ts
ruifigueira: Actually, that was a mistake of mine, I'm going to rollback it.

src/playwrightTestCLI.ts
ruifigueira: I was doing it that way because I was only checking if the trace viewer supported the playwright version in the extension, so I needed to send both implementations.

I'm now changing that to ensure that we only request the enabled trace viewer when needed, and so checking the playwright version can be done here or in testModel, so that we can only pass the right trace viewer to the call site.

src/settingsView.ts
```diff
@@ -18,6 +18,7 @@ import { DisposableBase } from './disposableBase';
 import type { ReusedBrowser } from './reusedBrowser';
 import type { SettingsModel } from './settingsModel';
 import type { TestModelCollection } from './testModel';
+import { getNonce } from './utils';
```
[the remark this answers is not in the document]
ruifigueira: PR for that: https://github.com/microsoft/playwright-vscode/pull/508

src/traceViewer.ts
ruifigueira: addressed in https://github.com/microsoft/playwright-vscode/pull/509

PR #491, opened 2024-05-22
fix: should not list empty workspace folders

mxschmitt:


src/testTree.ts
```diff
@@ -99,15 +89,20 @@ export class TestTree extends DisposableBase {
...
+      for (const model of this._models.enabledModels().filter(m => m.config.workspaceFolder === workspaceFolder.uri.fsPath)) {
         for (const project of model.enabledProjects())
           rootSuite.suites.push(project.suite as TeleSuite);
       }
-      const upstreamTree = new upstream.TestTree(workspaceFolder, rootSuite, [], undefined, path.sep);
+      const upstreamTree = new upstream.TestTree(workspaceFolder.uri.fsPath, rootSuite, [], undefined, path.sep);
       upstreamTree.sortAndPropagateStatus();
       upstreamTree.flattenForSingleProject();
+      if (upstreamTree.rootItem.children.length === 0) {
+        this._testController.items.delete(this._idWithGeneration(workspaceFolder.uri.fsPath));
```
pavelfeldman, line 102: Does delete work fine if the element is not there?

src/testTree.ts
```diff
@@ -99,15 +89,20 @@ export class TestTree extends DisposableBase {
   }
 
   private _update() {
-    for (const [workspaceFolder, workspaceRootItem] of this._rootItems) {
+    for (const workspaceFolder of this._vscode.workspace.workspaceFolders ?? []) {
```
pavelfeldman, line 92: I think the folder might be stuck if user removes a workspaceFolder.

src/testTree.ts
mxschmitt: Yup!

src/testTree.ts
mxschmitt: Addressed.

pavelfeldman (approved):

PR #498, opened 2024-06-14
bug(trace-viewer): update trace file on trace viewer when test completes

ruifigueira:
Fixes https://github.com/microsoft/playwright/issues/31314

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -488,10 +488,10 @@ export class Extension implements RunHooks {
           return;
 
         const trace = result.attachments.find(a => a.name === 'trace')?.path || '';
+        // if trace viewer is currently displaying the trace file about to be replaced, it needs to be refreshed
         const prevTrace = (testItem as any)[traceUrlSymbol];
         (testItem as any)[traceUrlSymbol] = trace;
-        if (enqueuedSingleTest || prevTrace === this._traceViewer.currentFile)
-          // refresh the trace viewer
+        if (enqueuedSingleTest || prevTrace === this._traceViewer.currentFile())
           this._showTrace(testItem);
 
         if (result.status === test.expectedStatus) {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -488,8 +488,10 @@ export class Extension implements RunHooks {
           return;
 
         const trace = result.attachments.find(a => a.name === 'trace')?.path || '';
+        const prevTrace = (testItem as any)[traceUrlSymbol];
         (testItem as any)[traceUrlSymbol] = trace;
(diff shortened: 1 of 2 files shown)


pavelfeldman:
I don't think the error that you see in console is related to the behavior that you are changing. What's the idea behind this change, could you verbalize it?

ruifigueira:
This change updates the trace displayed in the trace viewer when a test in a multiple test run completes (basically, it replaces the live json trace file, which is being polled during live mode, into the final zip trace file). Otherwise, VSCode `TraceViewer` instance doesn't send the new trace file to `_traceViewerProcess`, which will continue to send `loadTraceRequested` messages via websocket to the traceviewer webapp with the old json trace path.

In my tests, it fixed the issue.

ruifigueira:
BTW, the reason why it works with one single test running and fails with more than one is because `enqueuedSingleTest` is `true`  and so it calls the `showTrace`:

https://github.com/microsoft/playwright-vscode/blob/main/src/extension.ts#L486-L487

The check I added there is just to ensure that the trace viewer is showing the trace for the right test item. If they are the same, the trace viewer must be displaying the old trace value that is now being updated.

src/extension.ts
```diff
@@ -488,8 +488,10 @@ export class Extension implements RunHooks {
           return;
 
         const trace = result.attachments.find(a => a.name === 'trace')?.path || '';
+        const prevTrace = (testItem as any)[traceUrlSymbol];
```
pavelfeldman, line 491: Relying on the previous file is a bit fragile, but given that there is no clear way to get currently selected tree item, I'm ok with it. Let's add a comment explaining what is going on.

src/traceViewer.ts
```diff
@@ -38,6 +39,10 @@ export class TraceViewer implements vscodeTypes.Disposable {
     }));
   }
 
+  get currentFile() {
```
pavelfeldman, line 42: We don't use getters, should be plain method.

pavelfeldman (approved):

pavelfeldman:
> Otherwise, VSCode TraceViewer instance doesn't send the new trace file to _traceViewerProcess, which will continue to send loadTraceRequested messages via websocket to the traceviewer webapp with the old json trace path

That's normal behavior for the case of live traces - we race with the trace viewer here, so it is Ok for the files to not be there. I understand that you are removing a case where update is hammering against the missing file, but the JSON error will still be happening in other cases. I'd suggest that you change the title of the PR to reflect the actual change.

pavelfeldman (approved):

PR #503, opened 2024-07-05
test(trace-viewer): testing harness for webview panels

ruifigueira:
See https://github.com/microsoft/playwright-vscode/pull/483#discussion_r1657638327

Refs: #483

### DIFF
--- a/tests/mock/vscode.ts
+++ b/tests/mock/vscode.ts
@@ -26,17 +26,47 @@ import { CancellationToken } from '../../src/vscodeTypes';
 
 export class Uri {
   scheme = 'file';
-  fsPath!: string;
+  authority = '';
+  path = '';
+  query = '';
+  fragment = '';
+  fsPath = '';
 
   static file(fsPath: string): Uri {
     const uri = new Uri();
     uri.fsPath = fsPath;
+    uri.path = fsPath;
     return uri;
   }
 
   static joinPath(base: Uri, ...args: string[]): Uri {
     return Uri.file(path.join(base.fsPath, ...args));
   }
+
+  static parse(value: string): Uri {
+    const { protocol, host, pathname, search, hash } = new URL(value);
+    const uri = new Uri();
+    uri.scheme = protocol.replace(/:$/, '');
+    uri.authority = host;
+    uri.path = pathname;
+    uri.query = search;
+    uri.fragment = hash;
+    return uri;
+  }
+
+  toString() {
+    const url = new URL(`${this.scheme}://${this.authority}${this.path}`);
+    if (this.query) url.search = this.query;
+    if (this.fragment) url.hash = this.fragment;
+    return url.toString();
+  }
+}
+
+export enum ColorThemeKind {
+  Light = 1,
+  Dark = 2,
+  HighContrast = 3,
+  HighContrastLight = 4
 }
 
 class Position {
@@ -155,7 +185,7 @@ export class WorkspaceFolder {
   }
 }
 
-class TestItem {
+export class TestItem {
   readonly children = this;
   readonly map = new Map<string, TestItem>();
   range: Range | undefined;
@@ -797,6 +827,7 @@ type HoverProvider = {
 export class VSCode {
   isUnderTest = true;
   CancellationTokenSource = CancellationTokenSource;
+  ColorThemeKind = ColorThemeKind;
   DiagnosticSeverity = DiagnosticSeverity;
   EventEmitter = EventEmitter;
   Location = Location;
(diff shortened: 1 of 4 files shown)


tests/mock/vscode.ts
```diff
@@ -1024,31 +1093,73 @@ export class VSCode {
...
-        }
-        initializedPage.evaluate((data: any) => {
-          const event = new globalThis.Event('message');
-          (event as any).data = data;
-          globalThis.dispatchEvent(event);
-        }, data).catch(() => {});
-      };
-      provider.resolveWebviewView({ webview, onDidChangeVisibility: () => disposable });
+        currHtml = newValue;
+        initializedPage?.reload().catch(() => {});
```
pavelfeldman, line 1136: This would have a race for the cases of undefined initializedPage. Are we reassigning html in our code? What scenario are you trying to test here?

tests/mock/vscode.ts
```diff
@@ -1024,31 +1093,73 @@ export class VSCode {
...
+  }
+
+  async webViewsByPanelType(viewType: string) {
+    return (await Promise.all(this._webViewsByPanelType.get(viewType) ?? []));
+  }
+
+  async singleWebViewByPanelType(viewType: string): Promise<Page> {
+    const webViews = this._webViewsByPanelType.get(viewType);
+    if (!webViews?.size) {
+      while (true) {
```
pavelfeldman, line 1114: This makes it look quite messy. I would either await all webviews being created upon extension activation in the test mode, or would introduce a promise that resolves when all the webviews load, and then would expose the list of webviews w/ types or a getter by type.

tests/mock/vscode.ts
ruifigueira: Before using `playwrightTestServer` I was reassigning html (it first displayed a loading spinner while lauching a new playwright server and then I was injecting a new html with the iframe), but with the new approach it's no longer needed. I'll remove it.

tests/mock/vscode.ts
ruifigueira: >  I would either await all webviews being created upon extension activation in the test mode

That function is not for `WebviewViewProvider` like `settingsView`, it's for `WebviewPanel` like the trace viewer embedded panel, and those are dynamic and asyncronous. One example of its usage, where one webview panel opens, closes, and a new webview panel opens again:

https://github.com/microsoft/playwright-vscode/blob/e782ce5e1fc9700419080bc847cd5099e6cb4fa7/tests/trace-viewer.spec.ts#L252-L268

Before I was using `expect.poll` on `webViewsByPanelType` before trying to get the webview because there's asynchronous calls that may delay the creation of the Trace Viewer panel (for instance, [vscode.env.asExternalUrl](https://code.visualstudio.com/api/references/vscode-api#env.asExternalUri)).

ruifigueira: I converted `singleWebViewByPanelType` into a test util function in 3505843, and there I simply perform a `expect.poll` to ensure there's a unique webview for a given panel type. I think it's a much simpler solution and achieves the same goal.

tests/mock/vscode.ts
```diff
@@ -918,6 +956,25 @@ export class VSCode {
...
       return disposable;
     };
+    this.window.createWebviewPanel = (viewType: string) => {
+      const { webview, pagePromise } = this._createWebviewAndPage();
+      const didDispose = new EventEmitter<void>();
+      const panel: any = {};
+      panel.onDidDispose = didDispose.event;
+      panel.webview = webview;
+      pagePromise.then(webview => {
+        webview.on('close', () => panel.dispose());
```
pavelfeldman, line 966: I don't see webview.on('close') in VS Code so this is your invention it seems. What are you trying to do here? I understand how panel.dispose() should close webview and page, but what is this one about?

tests/mock/vscode.ts
```diff
@@ -918,6 +956,25 @@ export class VSCode {
...
+      });
+      const webviews = this._webViewsByPanelType.get(viewType) ?? new Set();
+      webviews.add(pagePromise);
```
pavelfeldman, line 969: I would only put resolved pages there to simplify the code.

tests/mock/vscode.ts
ruifigueira: It's to simulate closing the webview panel manually. I use it here

https://github.com/microsoft/playwright-vscode/blob/2c1d2e7972c1944926ff92a7daf21a5f955bf81e/tests/trace-viewer.spec.ts#L119

pavelfeldman (approved):

PR #509, opened 2024-07-25
feat(trace-viewer): refactor trace viewer

ruifigueira:
Refactoring before introducing embedded mode.

- TraceViewer renamed to SpawnTraceViewer and moved into testModel (see https://github.com/microsoft/playwright-vscode/pull/483#discussion_r1661655108)
- TestModel is now disposable, to ensure all resources are properly released when extension is disposed

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -759,6 +759,10 @@ export class Extension implements RunHooks {
     this._treeItemSelected(testItem);
   }
 
+  traceViewerInfoForTest() {
+    return this._traceViewer()?.infoForTest();
+  }
+
   private _showTrace(testItem: vscodeTypes.TestItem) {
     const traceUrl = (testItem as any)[traceUrlSymbol];
     if (traceUrl)
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -27,7 +27,6 @@ import { TestTree } from './testTree';
 import { NodeJSNotFoundError, ansiToHtml, getPlaywrightInfo } from './utils';
 import * as vscodeTypes from './vscodeTypes';
 import { WorkspaceChange, WorkspaceObserver } from './workspaceObserver';
-import { TraceViewer } from './traceViewer';
 import { registerTerminalLinkProvider } from './terminalLinkProvider';
 import { RunHooks, TestConfig } from './playwrightTestTypes';
 
@@ -67,7 +66,6 @@ export class Extension implements RunHooks {
   private _debugHighlight: DebugHighlight;
   private _isUnderTest: boolean;
   private _reusedBrowser: ReusedBrowser;
-  private _currentTraceViewer?: TraceViewer;
   private _settingsModel: SettingsModel;
   private _settingsView!: SettingsView;
   private _diagnostics: vscodeTypes.DiagnosticCollection;
@@ -237,6 +235,13 @@ export class Extension implements RunHooks {
       this._reusedBrowser,
       this._diagnostics,
       this._treeItemObserver,
+      this._models.onUpdated(() => {
+        const selectedModel = this._models.selectedModel();
+        for (const model of this._models.models()) {
+          if (model !== selectedModel)
+            model.traceViewer()?.close();
+        }
+      }),
       registerTerminalLinkProvider(this._vscode),
     ];
     const fileSystemWatchers = [
@@ -485,7 +490,7 @@ export class Extension implements RunHooks {
         // if trace viewer is currently displaying the trace file about to be replaced, it needs to be refreshed
         const prevTrace = (testItem as any)[traceUrlSymbol];
         (testItem as any)[traceUrlSymbol] = trace;
-        if (enqueuedSingleTest || prevTrace === this._currentTraceViewer?.currentFile())
+        if (enqueuedSingleTest || prevTrace === this._traceViewer()?.currentFile())
           this._showTrace(testItem);
 
         if (result.status === test.expectedStatus) {
@@ -537,7 +542,7 @@ export class Extension implements RunHooks {
     if (isDebug) {
       await model.debugTests(items, testListener, testRun.token);
     } else {
-      await this._enabledTraceView()?.willRunTests();
+      await this._traceViewer()?.willRunTests();
       await model.runTests(items, testListener, testRun.token);
     }
   }
@@ -757,24 +762,20 @@ export class Extension implements RunHooks {
   private _showTrace(testItem: vscodeTypes.TestItem) {
     const traceUrl = (testItem as any)[traceUrlSymbol];
     if (traceUrl)
-      this._enabledTraceView()?.open(traceUrl);
+      this._traceViewer()?.open(traceUrl);
   }
 
   private _treeItemSelected(treeItem: vscodeTypes.TreeItem | null) {
     if (!treeItem)
       return;
     const traceUrl = (treeItem as any)[traceUrlSymbol] || '';
-    if (!traceUrl && !this._currentTraceViewer?.isStarted())
+    if (!traceUrl && !this._traceViewer()?.isStarted())
       return;
-    this._enabledTraceView()?.open(traceUrl);
+    this._traceViewer()?.open(traceUrl);
   }
 
-  private _enabledTraceView() {
-    const traceViewer = this._models.selectedModel()?.enabledTraceViewer();
-    if (traceViewer !== this._currentTraceViewer)
-      this._currentTraceViewer?.close();
-    this._currentTraceViewer = traceViewer;
-    return traceViewer;
+  private _traceViewer() {
+    return this._models.selectedModel()?.traceViewer();
   }
 
   private _queueCommand<T>(callback: () => Promise<T>, defaultValue: T): Promise<T> {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -235,13 +235,6 @@ export class Extension implements RunHooks {
       this._reusedBrowser,
       this._diagnostics,
       this._treeItemObserver,
-      this._models.onUpdated(() => {
-        const selectedModel = this._models.selectedModel();
-        for (const model of this._models.models()) {
-          if (model !== selectedModel)
-            model.traceViewer()?.close();
-        }
-      }),
       registerTerminalLinkProvider(this._vscode),
     ];
     const fileSystemWatchers = [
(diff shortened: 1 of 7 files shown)


src/extension.ts
```diff
@@ -756,18 +756,25 @@ export class Extension implements RunHooks {
...
-      this._traceViewer.open(traceUrl, testModel.config);
+    if (!traceUrl && !this._currentTraceViewer?.isStarted())
+      return;
+    this._enabledTraceView()?.open(traceUrl);
+  }
+
+  private _enabledTraceView() {
+    const traceViewer = this._models.selectedModel()?.enabledTraceViewer();
+    if (traceViewer !== this._currentTraceViewer)
+      this._currentTraceViewer?.close();
```
pavelfeldman, line 775: I wasn't expecting enabledTraceViewer to do something. Also the naming is probably suboptimal. If `enabledTraceViewer()` "closes" trace viewers, we'll run of trace viewers. Does it mean to say closeTrace?

src/testModel.ts
```diff
@@ -89,6 +92,22 @@ export class TestModel {
...
+    this._disposables.push(
+        this._spawnTraceViewer,
+        options.settingsModel.showTrace.onChange(value => {
+          if (!value)
+            this._spawnTraceViewer.close();
+        }),
+    );
+  }
+
+  enabledTraceViewer() {
```
pavelfeldman, line 106: I'd call it traceViewer() at this point.

src/testModel.ts
```diff
@@ -665,7 +690,10 @@ export class TestModelCollection extends DisposableBase {
     const configSettings = (workspaceSettings.configs || []).find(c => c.relativeConfigFile === path.relative(model.config.workspaceFolder, model.config.configFile));
     model.isEnabled = configSettings?.enabled || (this._models.length === 1 && !configSettings);
     await this._loadModelIfNeeded(model);
-    this._disposables.push(model.onUpdated(() => this._didUpdate.fire()));
+    this._disposables.push(
+        model,
```
pavelfeldman, line 694: `TestModelCollection.clear()` will call both dispose() and reset() now. Also you might want to hide model's trace viewer when model is being reset (that happens when it is enabled and disabled, it is worth closing trace viewer on that signal.

ruifigueira: > TestModelCollection.clear() will call both dispose() and reset() now

You're right, I'll change it to ensure it only calls `testModel.reset`, otherwise we'll end up with some nasty bugs.

> Also you might want to hide model's trace viewer when model is being reset

Trace viewer already closes when model is reset:

https://github.com/microsoft/playwright-vscode/blob/f01b9d47487d2d93091d8305aa5912e28455445d/src/testModel.ts#L124

However, it doesn't close when another test model is selected. I'll implement that.

ruifigueira: Actually,, `TestModelCollection.clear()` actually disposes testModels, so I think the current behaviour is right. If it only called `reset`, `options.settingsModel.showTrace.onChange` event handler would not be disposed.

src/testModel.ts
ruifigueira: Addressed in https://github.com/microsoft/playwright-vscode/pull/509/commits/643d3c37a4d9c71e7d12ecb2e56016e98feccfe9

src/extension.ts
ruifigueira: Ok, I understand. Instead of closing trace viewer here, I'll make sure trace viewers corresponding to unselected testModels are closed.

Addressed in https://github.com/microsoft/playwright-vscode/pull/509/commits/643d3c37a4d9c71e7d12ecb2e56016e98feccfe9.

src/extension.ts
```diff
@@ -228,6 +225,14 @@ export class Extension implements RunHooks {
       this._reusedBrowser,
       this._diagnostics,
       this._treeItemObserver,
+      this._models,
+      this._models.onUpdated(() => {
```
pavelfeldman, line 229: Looks like this code operates `this._models` and its content, so it should be a part of `TestModelCollection`. It can probably be handled more efficiently there.

src/testModel.ts
pavelfeldman: Ok, sounds good. We should probably follow up with getting rid of

```js
clear() {
  this.dispose();
}
```

for clarity. `clear` assumes that the instance will be reused. But if in the future someone adds an essential listener into `disposables` in TestModelCollection, calling `clear()` will be destroying it and hence leading to bugs in TestModelCollection.

src/traceViewer.ts
```diff
@@ -95,26 +81,48 @@ export class TraceViewer implements vscodeTypes.Disposable {
...
     traceViewerProcess.on('error', error => {
       this._vscode.window.showErrorMessage(error.message);
-      this.close().catch(() => {});
+      this.close();
     });
+    if (this._vscode.isUnderTest) {
+      traceViewerProcess.stdout?.on('data', data => {
+        const match = data.toString().match(/Listening on (.*)/);
+        if (match)
+          this._serverUrlPrefix = match[1];
```
pavelfeldman, line 90: If this is only used for test, you should mention it in the variable name.

tests/spawn-trace-viewer.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  await expect.poll(() => traceViewerInfo(vscode)).toBeUndefined();
+});
+
+test('should refresh trace viewer while test is running', async ({ activate }) => {
+  const { vscode, testController } = await activate({
+    'playwright.config.js': `module.exports = { testDir: 'tests' }`,
+    'tests/test.spec.ts': `
+      import { test } from '@playwright/test';
+      test('should pass', async () => await new Promise(r => setTimeout(r, 1000)));
```
pavelfeldman, line 94: 1000 won't be enough on CI. It is best to implement a latch here, we have it in ui mode test harness, so you can copy and reuse it here: https://github.com/microsoft/playwright/blob/main/tests/playwright-test/ui-mode-test-progress.spec.ts#L124

tests/spawn-trace-viewer.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  await Promise.all([
+    testController.run(),
+    expect.poll(() => traceViewerInfo(vscode)).toMatchObject({
+      type: 'spawn',
+      traceFile: expect.stringMatching(/\.json$/),
+    }),
+  ]);
+
+  await expect.poll(() => traceViewerInfo(vscode)).toMatchObject({
```
pavelfeldman, line 109: You'll be opening the latch before this line

pavelfeldman (commented):
Love the test coverage!

src/testModel.ts
ruifigueira: Maybe to make it clear:

- instead of extend `DisposableBase`, just implement `vscodeType.Disposable`
- intead of `_disposables`, keep model disposables in a `private _modelsDisposables: Disposable[];`
- implement `clear` and `dispose` functions this way:

```js
  clear() {
    this._disposeModels();
    this._didUpdate.fire();
  }

  dispose() {
    this._disposeModels();
  }

  private _disposeModels() {
    for (const disposable of this.__modelsDisposables)
     disposable.dispose();
    this._models = [];
  }
```

ruifigueira: Actually, to move the `onUpdated` inside of `TestModelCollection` we still need to keep the `DisposableBase` and keep the additional `_modelsDisposables` property to release cleared models. I applied those changes in b8551f60a0b309ca1627399fe060490ddcb910d2

src/testModel.ts
```diff
@@ -615,12 +640,22 @@ export class TestModelCollection extends DisposableBase {
   private _didUpdate: vscodeTypes.EventEmitter<void>;
   readonly onUpdated: vscodeTypes.Event<void>;
   private _context: vscodeTypes.ExtensionContext;
+  private _modelsDisposables: vscodeTypes.Disposable[] = [];
```
pavelfeldman, line 643: This is still quite confusing, sorry. We should be able to handle this use case without additional complexity. All the objects we operate have nested lifetimes:

- TestModelCollection outlives models that it contains
- TestModels contain respective trace viewers
- There is a way to clear the collection that destroys all the models due to containment
- There is a way to reset the model while maintaining it alive.

I'm not sure why we can't keep things straightforward, but maybe it is the view aspect that makes us keep just one trace viewer open that stands in the way?

ruifigueira: Current `TestModelCollection` in main branch is not included in the extensions disposables, and I needed to add it to ensure it was properly released in case the extension is deactivated. But `TestModelCollection` dispose in main branch doesn't reset its models, only disposes the `model.onUpdate` event handlers (it uses the `DisposableBase` implementation).

So, to make this step clear, maybe I can do another PR just trying to fix that.

ruifigueira: Done: https://github.com/microsoft/playwright-vscode/pull/511

ruifigueira: Sometimes it's better to move one step back: if we keep trace viewer instances in settings we can avoid registering event handlers in `TestModelCollection` and `TestModel`, so basically I'm rolling back changes in those classes.

So:

- TraceViewer implementations become shallow objects once again, with the same lifetime as the extension (right now, it's only `SpawnTraceViewer`, but the same logic will be applied to `EmbeddedTraceViewer` in the next phase)
- They keep a reference to `TestModelCollection` so that they can get the current configuration / server URL
- All event handling that can impact trace viewers (`models.onUpdated` / `showTrace.onChange`, `embeddedTraceViewer.onChange`) are handled in extensions

Let me know if this approach is better

pavelfeldman: It is hard for me to give a good advice without doing the actual work. My naive thought process is as follows:
- User selects test item in the tree
- Extension (controller) handles this change and makes a decision that trace for the test result should be shown
- Extension creates (or reuses) trace viewer for this test via the test model.
- Extension takes owership of the trace viewer and listens for the model to be disabled and/or removed, destroys dangling trace viewer when needed.
- Extension opens trace in the created trace viewer

ruifigueira: Ok my latest commit no longer makes sense with the changes you introduced. I'll drop it.

pavelfeldman: I don't think I changed anything fundamentally, but you would know better what applies and what does not. Please let me know which patch I should be reviewing!

ruifigueira: you can review the current one, I already rolled back the "step back" solution.

pavelfeldman (approved):

PR #510, opened 2024-07-30
Convert mocked VSCode tests into integration tests PoC

ruifigueira:
I created a small test framework on top of [VSCode extension testing runner](https://code.visualstudio.com/api/working-with-extensions/testing-extension#advanced-setup-your-own-runner) and playwright tests, extending the existing [tests-integration](https://github.com/microsoft/playwright-vscode/blob/main/tests-integration) suite.

The main idea is to be able to evaluate VSCode-side code from playwright tests, in a similar way as [Page.evaluate / Page.evaluateHandle](https://playwright.dev/docs/evaluating). Besides, I also introduced the `VSCodeHandle` to keep remote references to VSCode objects that those functions may return, as well as registering local listeners over remote `EventEmitter`s.

Another nice thing with this approach is that using a library as `sinon` (I use it to decorate some vscode behaviour) is as simple as installing it with `npm install` and dynamically import it in the VSCode function to evaluate.

With all this, I was able to create proxies with similar API to the existing mocked VSCode and converting test specs to use the integration library is as easy as changing the import, basically.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
         "@types/babel__traverse": "^7.20.3",
         "@types/glob": "^8.0.0",
         "@types/node": "^18.11.9",
+        "@types/sinon": "^17.0.3",
         "@types/stack-utils": "^2.0.1",
         "@types/vscode": "1.86.0",
         "@types/which": "^3.0.3",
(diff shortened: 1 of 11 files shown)


pavelfeldman:
Could you help me understand how you intend to be using this? We already have a test harness that scales well enough to allow for large number of tests with options matrices. The idea behind the integration tests is that extension does not fail miserably in the production, they were not intended to be used for correctness testing.

ruifigueira:
It was just an exercise, to see if it was feasible.

When I was extending the mocked vscode for the embedded trace viewer, I wondered if it was possible to use vscode itself. I then realized that you already had those integration tests, and so I just tried to fill the gaps.

pavelfeldman:
I'll close it as outside of the scope.

PR #511, opened 2024-07-31
fix: ensure TestModelCollection is disposed

ruifigueira:


src/testModel.ts
```diff
@@ -709,10 +709,14 @@ export class TestModelCollection extends DisposableBase {
 
   clear() {
     this.dispose();
```
pavelfeldman, line 711: I was suggesting that this is the actual problem, i.e. no method in the class should be calling `dispose()` since after calling dispose, we can't use that instance. Imagine one of the disposables is essential to the instance operation, now clearing its state will be corrupting the instance.

ruifigueira: I understand, `dispose` should call `clear`, not this way around. I think that extending `TestModelCollection` extending `DisposableBase` in this case doesn't help, because it implicitly implements the `dispose` method the wrong way, and to make it right we need to call `super.dispose()` in `clear` and then `this.clear()` in the overriden `dispose` method.

ruifigueira: Actually, dispose cannot call `this.clear()` because it would fire `onUdate` event. That's why I added the `_disposeModels` function in one of my attempts, which was then called in both `clear` and `dispose` functions

pavelfeldman:
I moved some code around in https://github.com/microsoft/playwright-vscode/pull/512/files, please feels free to review. It should streamline clear vs dispose for you.

ruifigueira:
I'm ok with those changes. I'll rebase my traceViewer refactoring branch against it.

PR #513, opened 2024-08-06
Feature: Introducing embedded Trace Viewer

ruifigueira:
See https://github.com/microsoft/playwright-vscode/pull/483#pullrequestreview-2195415779


package.json
```diff
@@ -70,6 +70,11 @@
         "command": "pw.extension.toggle.showTrace",
         "title": "%contributes.command.pw.extension.toggle.showTrace%"
       },
+      {
```
pavelfeldman, line 73: Let's keep it as a hidden setting at first, not visible in a sidebar, no toggle action.

package.json
```diff
@@ -113,6 +118,11 @@
           "type": "boolean",
           "default": false,
           "description": "%configuration.playwright.showTrace%"
+        },
+        "playwright.embeddedTraceViewer": {
```
pavelfeldman, line 122: I.e. we will only have this.

src/settingsView.ts
```diff
@@ -278,6 +278,10 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
             <input type="checkbox" setting="showTrace"></input>
             ${vscode.l10n.t('Show trace viewer')}
           </label>
+          <label>
```
pavelfeldman, line 281: Remove this

src/embeddedTraceViewer.ts
```diff
@@ -0,0 +1,260 @@
...
+
+export class EmbeddedTraceViewer implements TraceViewer {
+  readonly vscode: vscodeTypes.VSCode;
+  readonly extensionUri: vscodeTypes.Uri;
+  private _settingsModel: SettingsModel;
+  private _currentFile?: string;
+  private _traceViewerPanel: EmbeddedTraceViewerPanel | undefined;
+  private _traceLoadRequestedTimeout?: NodeJS.Timeout;
+  private _config: TestConfig;
+  private _testServer?: PlaywrightTestServer;
```
pavelfeldman, line 34: I don't think embedded trace viewer makes sense w/o the test server, make it required?

src/embeddedTraceViewer.ts
```diff
@@ -0,0 +1,260 @@
...
+    const serverUrl = new URL(wsEndpoint);
+    serverUrl.protocol = 'http';
+    if (this.vscode.env.remoteName && ['[::1]', '0.0.0.0'].includes(serverUrl.hostname))
+      serverUrl.hostname = 'localhost';
+    const serverUri = await this.vscode.env.asExternalUri(this.vscode.Uri.parse(serverUrl.origin));
+    return serverUri.toString().replace(/\/$/, '');
+  }
+
+  private async _startIfNeeded() {
+    if (this._traceViewerPanel)
```
pavelfeldman, line 101: This code has inherent race condition where calling it 5 times will create 5 panel instances.

src/embeddedTraceViewer.ts
```diff
@@ -0,0 +1,260 @@
...
+    });
+    this._webviewPanel.iconPath = this._vscode.Uri.joinPath(this._extensionUri, 'images', 'playwright-logo.svg');
+    this._webviewPanel.webview.html = this._getHtml();
+    this._disposables = [
+      this._webviewPanel,
+      this._webviewPanel.onDidDispose(() => {
+        embeddedTestViewer.close();
+      }),
+      this._webviewPanel.webview.onDidReceiveMessage(({ command, params }) => {
+        if (typeof command === 'string')
```
pavelfeldman, line 170: What else are we expecting? Is it safe to destructure it one line above?

src/playwrightTestServer.ts
```diff
@@ -43,6 +44,12 @@ export class PlaywrightTestServer {
     this._disposeTestServer();
   }
 
+  async wsEndpoint() {
```
pavelfeldman, line 47: It is unclear this starts the server. Let's make it more clear and specific: ensureStartedForTraceViewer(): and return trace viewer URL

src/testModel.ts
```diff
@@ -91,24 +93,24 @@ export class TestModel extends DisposableBase {
...
-    if (this._spawnTraceViewer.checkVersion())
-      return this._spawnTraceViewer;
-  }
+    if (this._currentTraceViewer !== undefined)
+      return this._currentTraceViewer;
 
-  _closeTraceViewerIfNeeded() {
-    if (this._collection.selectedModel() !== this || !this._embedder.settingsModel.showTrace.get())
-      this._spawnTraceViewer.close();
+    this._currentTraceViewer = this._traceViewers.find(t => t.checkSupport(true)) ?? null;
```
pavelfeldman, line 112: It does not look like this is necessarily a user gesture. Looks like a generic programmatic getter to me

src/testModel.ts
```diff
@@ -383,6 +386,17 @@ export class TestModel extends DisposableBase {
     return this._filesPendingListTests.promise;
   }
 
+  private _closeTraceViewerIfNeeded() {
+    if (
+      this._collection.selectedModel() !== this ||
+      // check if there's a high priority trace viewer that supports currennt configuration
+      this._traceViewers.find(t => t.checkSupport(false)) !== this._currentTraceViewer
+    ) {
+      this._currentTraceViewer?.close();
```
pavelfeldman, line 395: The name does not make sense to me - I think once you closed the trace viewer you can't use it. So either traceViewer should be a traceViewerFactory, or this should be closeTraceView().

src/traceViewer.ts
```diff
@@ -18,21 +18,37 @@ import { ChildProcess, spawn } from 'child_process';
...
   private _vscode: vscodeTypes.VSCode;
+  private _settingsModel: SettingsModel;
   private _envProvider: () => NodeJS.ProcessEnv;
   private _traceViewerProcess: ChildProcess | undefined;
   private _currentFile?: string;
   private _config: TestConfig;
   private _serverUrlPrefixForTest?: string;
 
-  constructor(vscode: vscodeTypes.VSCode, envProvider: () => NodeJS.ProcessEnv, config: TestConfig) {
+  constructor(vscode: vscodeTypes.VSCode, settingsModel: SettingsModel, envProvider: () => NodeJS.ProcessEnv, config: TestConfig) {
```
pavelfeldman, line 47: nit: Passing settingsModel does not seem justified atm, but I'm fine with it

tests-integration/tests/baseTest.ts
```diff
@@ -68,15 +80,30 @@ export const test = base.extend<TestFixtures>({
       await fs.promises.cp(logPath, logOutputPath, { recursive: true });
     }
   },
-  createProject: async ({ createTempDir }, use) => {
+  getWebview: async ({ workbox }, use) => {
```
pavelfeldman, line 83: I'm not reviewing this as a part of this change, could you please extract it and explain what and why is happening here?

tests/utils.ts
```diff
@@ -37,15 +37,14 @@ type Latch = {
 type TestFixtures = {
```
pavelfeldman, line 37: Let's land the testing harness changes separately, before this change.

pavelfeldman (commented):
Looks great, thanks!

pavelfeldman:
I also noticed a bunch of bugs, but we don't need to fix them right now. Selection seems to be broken and dragging the view between tabs does not restore its state (this one will be interesting)

src/embeddedTraceViewer.ts
ruifigueira: I was assuming that the **Embedded** checkbox would be available even in legacy versions that had `PlaywrightTestCLI`, and that way the `EmbeddedTraceViewer` instance would show the warning message.

But if the warning message is not needed, I can make it required and only instanciate `EmbeddedTraceViewer` when it makes sense. In that case, I can also move the `checkSupport` logic into `testModel` and remove `settingsModel` dependency from trace viewer.

src/traceViewer.ts
ruifigueira: I think changes I'll do on https://github.com/microsoft/playwright-vscode/pull/513#discussion_r1706132388 will no longer require this dependency on `settingsModel`

tests-integration/tests/baseTest.ts
ruifigueira: I'll remove the integration test commit from this PR, and I'll include it after these merges are done.

tests/utils.ts
ruifigueira: done: https://github.com/microsoft/playwright-vscode/pull/514

src/testModel.ts
ruifigueira: I was keeping the same pattern from `TraceViewer` before my changes: one instance with the same lifecycle as its container (before it was `extension`, now it's `testModel`) that could be closed and open several times.

And the idea of this `_closeTraceViewerIfNeeded` is to only close the trace viewer when either the selected model actually changes or when current trace viewer becomes due to settings changes (e.g., checking the embedded setting would close the spawn trace viewer and open the embedded),

There are situations where changing trace related settings won't close the current trace viewer. For instance, if it currently has embedded checked but playwright version doesn't support it, it will open spawn trace viewer. If the user then unchecks the embedded setting, it should keep the spawn trace viewer open.

src/embeddedTraceViewer.ts
ruifigueira: Messages that get here are from `playwright/packages/trace-viewer` module.

I was reviewing the changes I did there and unfortunatelly I used different message keys, both `command` and `message`:

- `command: 'openExternal'`: 

https://github.com/microsoft/playwright/blob/7ec3a93db3607e39b001e700020635bb017b4508/packages/trace-viewer/src/ui/embeddedWorkbenchLoader.tsx#L26

- `method: 'showErrorMessage'`:

https://github.com/microsoft/playwright/blob/7ec3a93db3607e39b001e700020635bb017b4508/packages/trace-viewer/src/ui/embeddedWorkbenchLoader.tsx#L82

So I need to change this, and I'll also create a new PR in playwright to make them consistent (with `method`, which is already very common in playwright code for these kind of messages).

ruifigueira:
I'm also experiencing some odd behaviour when launching tests in debug. From what I can tell, when debugging from vscode, it deactivates tracing:

https://github.com/microsoft/playwright-vscode/blob/e94a82316ca54e45f27025e5f44908a344b8f1a4/src/testModel.ts#L397

but nevertheless, it sets the live json path in the test item:

https://github.com/microsoft/playwright-vscode/blob/98c19d608e0983b8086de42dace7fff53c626b10/src/extension.ts#L461-L462

This is causing the trace viewer to open in debug, without displaying any trace because the path is not valid.

ruifigueira:
> dragging the view between tabs does not restore its state

This option does the trick: https://code.visualstudio.com/api/extension-guides/webview#retaincontextwhenhidden .

It doesn't work when moving the panel into a new window, though.

src/embeddedTraceViewer.ts
pavelfeldman: Yes, let's only support Embedded for modern PlaywrightTestServer-based configurations only.

src/testModel.ts
pavelfeldman: Let's land this bit as is and if I come up with a better name I'll follow up.

pavelfeldman:
> This is causing the trace viewer to open in debug, without displaying any trace because the path is not valid.

If this bug exists on tip of the tree we should file and fix it

ruifigueira:
I already open a PR to fix that bug: https://github.com/microsoft/playwright-vscode/pull/515

src/embeddedTraceViewer.ts
ruifigueira: Addressed in 31a15f3f926c5f17a8e16a4476d52c62f11778a9

src/embeddedTraceViewer.ts
ruifigueira: Addressed in 7b384dbe30b8ccc1fde6d310adb668c8a9421e65

src/playwrightTestServer.ts
ruifigueira: Addressed in 49d717f955696790e6d61abb7e91cf29ba63afc1

src/testModel.ts
ruifigueira: Addressed in 31a15f3f926c5f17a8e16a4476d52c62f11778a9 with `traceViewer()` and `ensureTraceViewer()`. The latter creates a trace viewer if not initialized, as a result of a user gesture (running tests or selecting a test item).

src/testModel.ts
ruifigueira: I left the `_closeTraceViewerIfNeeded` function but created a traceViewerFactory in 31a15f3f926c5f17a8e16a4476d52c62f11778a9 to check which trace viewer strategy to instanciate.

src/traceViewer.ts
ruifigueira: Addressed in 31a15f3f926c5f17a8e16a4476d52c62f11778a9

src/testModel.ts
```diff
@@ -57,6 +58,64 @@ type Watch = {
   include: readonly vscodeTypes.TestItem[] | undefined;
 };
 
+class TraceViewerFactory {
```
pavelfeldman, line 61: Move this into its own file?

src/traceViewer.ts
```diff
@@ -19,9 +19,20 @@ import type { TestConfig } from './playwrightTestTypes';
...
+  infoForTest(): {
+    type: string;
+    serverUrlPrefix?: string;
+    testConfigFile: string;
+    traceFile?: string;
+  } | undefined;
+};
 
-export class SpawnTraceViewer {
+export class SpawnTraceViewer implements TraceViewer {
```
pavelfeldman, line 35: Move this out into its own file as a separate change and then you can place TraceViewerFactory here.

src/embeddedTraceViewer.ts
```diff
@@ -0,0 +1,263 @@
...
+    this._currentFile = undefined;
+  }
+
+  private async _startIfNeeded() {
+    if (this._testServerStartedPromise)
+      return;
+
+    this._testServerStartedPromise = this._testServer.ensureStartedForTraceViewer();
+    const serverUrlPrefix = await this._testServerStartedPromise;
+    // if undefined, it means it was closed while test server started
```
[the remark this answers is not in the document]
ruifigueira: It was my initial thought, but I couldn't dispose `traceViewerPanel` synchronously and forces `infoForTest` to become async. But I'll do it that way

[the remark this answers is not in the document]
ruifigueira: Addressed in 1942a463a88da42e1ba8a1d4628c57476eaa4ff6

src/embeddedTraceViewer.ts
```diff
@@ -0,0 +1,263 @@
...
+  close() {
+    this._traceViewerPanel?.dispose();
+    this._traceViewerPanel = undefined;
+    this._testServerStartedPromise = undefined;
+    this._currentFile = undefined;
+  }
+
+  private async _startIfNeeded() {
+    if (this._testServerStartedPromise)
+      return;
```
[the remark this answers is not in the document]
ruifigueira: Addressed in 1942a463a88da42e1ba8a1d4628c57476eaa4ff6

src/traceViewer.ts
ruifigueira: Addressed in 87543434f44c7494c7e7e46678f88bd4173db83a

package.nls.json
```diff
@@ -14,6 +14,7 @@
   "contributes.command.pw.extension.command.toggleModels": "Toggle Playwright configs",
   "contributes.command.pw.extension.toggle.reuseBrowser": "Toggle Playwright browser reuse",
   "contributes.command.pw.extension.toggle.showTrace": "Toggle Playwright Trace Viewer",
+  "contributes.command.pw.extension.toggle.embeddedTraceViewer": "Toggle Embedded Playwright Trace Viewer",
```
pavelfeldman, line 17:
```suggestion
  "configuration.playwright.embeddedTraceViewer": "Toggle Embedded Playwright Trace Viewer",
```

l10n/bundle.l10n.json
```diff
@@ -18,9 +18,10 @@
   "Show browser mode does not work in remote vscode": "Show browser mode does not work in remote vscode",
   "Show browser": "Show browser",
   "Show trace viewer": "Show trace viewer",
+  "embedded trace viewer": "embedded trace viewer",
```
pavelfeldman, line 21: Unused, revert

src/testModel.ts
```diff
@@ -89,26 +90,38 @@ export class TestModel extends DisposableBase {
...
-      this._spawnTraceViewer.close();
+  ensureTraceViewer(): TraceViewer | null {
+    if (this._traceViewer === undefined)
+      this._traceViewer = this._traceViewerFactory.create();
+
+    return this._traceViewer;
+  }
+
+  private _closeTraceViewerIfNeeded() {
+    if (this._traceViewer && this._collection.selectedModel() === this && this._traceViewerFactory.isSupported(this._traceViewer))
```
pavelfeldman, line 121: This will never close spawned trace viewer since it is always supported.

src/testModel.ts
```diff
@@ -89,26 +90,38 @@ export class TestModel extends DisposableBase {
...
+      this._traceViewerFactory = new TraceViewerFactory(collection, this.config);
+    } else {
+      this._playwrightTest = new PlaywrightTestServer(this._vscode, this, collection.embedder);
+      this._traceViewerFactory = new TraceViewerFactory(collection, this.config, this._playwrightTest);
+    }
     this.tag = new this._vscode.TestTag(this.config.configFile);
-    this._spawnTraceViewer = new SpawnTraceViewer(this._vscode, this._embedder.envProvider, this.config);
 
     this._disposables = [
       this._embedder.settingsModel.showTrace.onChange(() => this._closeTraceViewerIfNeeded()),
```
pavelfeldman, line 103: For the embedded version, I expect this toggle to show the trace viewer upon enabling.

src/embeddedTraceViewer.ts
```diff
@@ -0,0 +1,261 @@
...
+  constructor(
+    embeddedTestViewer: EmbeddedTraceViewer,
+    serverUrlPrefix: string
+  ) {
+    super();
+    this._vscode = embeddedTestViewer.vscode;
+    this._extensionUri = embeddedTestViewer.extensionUri;
+    this.serverUrlPrefix = serverUrlPrefix;
+    this._isVisible = false;
+    this._webviewPanel = this._vscode.window.createWebviewPanel(EmbeddedTraceViewerPanel.viewType, 'Trace Viewer', {
```
pavelfeldman, line 107: Since we are not implementing an editor of the selected document, we should not use this API to show trace viewer. We should instead show it as a sidebar or a drawer panel and encourage user to drag it to the right. Try doing it with the Playwright settings and you'll see what I mean.

pavelfeldman (approved):
The change looks great. It still has several small problems, but I am open to landing it as is as long as you'll follow up to these issues!

l10n/bundle.l10n.json
ruifigueira: Addressed in b8a3d8b105bfd17b1c4c1dd8b4263723356ef219

package.nls.json
ruifigueira: Addressed in b8a3d8b105bfd17b1c4c1dd8b4263723356ef219

src/testModel.ts
ruifigueira: Addressed in eaf9074019931773ad149d158711f9184846c4db

src/embeddedTraceViewer.ts
ruifigueira: Do you mean this?

image

I honestly don't really like that approach. I'd rather go with editors. I already started to implement it but I left it on hold while this PR was under review.

For reference, here's a screenshot of the current status, with two trace editors side by side:

image

I did use the `vscode.window.createWebviewPanel` API because it's the one that offers a similar user experience to editors.

l10n/bundle.l10n.json
```diff
@@ -23,4 +23,4 @@
   "this feature": "this feature",
   "Toggle Playwright Configs": "Toggle Playwright Configs",
   "When enabled, Playwright will reuse the browser instance between tests. This will disable parallel execution.": "When enabled, Playwright will reuse the browser instance between tests. This will disable parallel execution."
-}
\ No newline at end of file
+}
```
pavelfeldman, line 26: why touching this file?

src/testModel.ts
pavelfeldman: I was referring to changing the settings value. If you change the setting value nothing will happen.

src/embeddedTraceViewer.ts
pavelfeldman: Traces side by side is a good example of why you should not be using the editor pattern. It is unclear what controls what. If trace viewer was opened for a trace file, it would be a different story.

pavelfeldman:
I landed the change because every time you change one line I need to review this whole huge PR, it is time consuming. Please make incremental changes that can be code-reviewed, otherwise we are at risk of compromising the product quality!

src/embeddedTraceViewer.ts
ruifigueira: I created URIs so that they got proper breadcrumbs.

Meanwhle, I was able to implement it with webviews. Here's a screenshot of it:

image

, I'll clean the code a bit and create a new PR for that when I have some spare time.

PR #515, opened 2024-08-08
[Bug] trace viewer opens in empty state while debugging

ruifigueira:
Bug detected in https://github.com/microsoft/playwright-vscode/pull/513#issuecomment-2274495198, but it also happens with public releases of playwright / playwright-vscode.

To reproduce:

- initialize a playwright test project with `npm init playwright@latest`
- in **Testing**, ensure **Show trace viewer** is selected
- in **Testing Explorer**, expand and select a leaf test item, e.g. `has title`
- click the **Debug Test** button

Eventually, it will open the trace viewer app showing **Select test to see the trace**.

This PR tries to fix it by ensuring that trace paths are only associated with corresponding test items when they are not running in debug.

## Environment:

```
  System:
    OS: Windows 11 10.0.22631
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
    Memory: 11.99 GB / 31.74 GB
  Binaries:
    Node: 20.15.0 - ~\scoop\apps\nvm\current\nodejs\nodejs\node.EXE
    npm: 10.7.0 - ~\scoop\apps\nvm\current\nodejs\nodejs\npm.CMD
    pnpm: 8.15.4 - ~\AppData\Local\pnpm\pnpm.EXE
  IDEs:
    VSCode: 1.92.0 - C:\Users\rui.figueira\scoop\apps\vscode\current\bin\code.CMD
  npmPackages:
    @playwright/test: ^1.46.0 => 1.46.0

Playwright Test for VSCode: v1.1.7
```

### DIFF
--- a/tests/trace-viewer.spec.ts
+++ b/tests/trace-viewer.spec.ts
@@ -15,9 +15,18 @@
  */
 
 import { enableConfigs, expect, selectConfig, selectTestItem, test, traceViewerInfo } from './utils';
+import { TestItem } from './mock/vscode';
 
 test.skip(({ showTrace }) => !showTrace);
-test.skip(({ showTrace, overridePlaywrightVersion }) => overridePlaywrightVersion && showTrace === 'embedded');
+
+async function waitForTestItemStatus(testItem: TestItem, status: TestItem['status']) {
+  return await new Promise<void>(resolve => testItem.testController.onDidCreateTestRun(testRun => {
+    testRun.onDidChange(() => {
+      if (testItem.status === status)
+        resolve();
+    });
+  }));
+}
 
 test('@smoke should open trace viewer', async ({ activate, showTrace }) => {
   const { vscode, testController } = await activate({
@@ -95,9 +104,12 @@ test('should refresh trace viewer while test is running', async ({ activate, cre
   });
 
   await testController.expandTestItems(/test.spec/);
-  selectTestItem(testController.findTestItems(/pass/)[0]);
+  const [testItem] = testController.findTestItems(/pass/);
+  selectTestItem(testItem);
 
   const testRunPromise = testController.run();
+  await waitForTestItemStatus(testItem, 'started');
+
   await expect.poll(() => traceViewerInfo(vscode)).toMatchObject({
     type: showTrace,
     traceFile: expect.stringMatching(/\.json$/),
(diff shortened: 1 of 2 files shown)


pavelfeldman:
I did not realize this was intended to land first.

ruifigueira:
I'll rebase it, no worries.

tests/trace-viewer.spec.ts
```diff
@@ -94,9 +102,13 @@ test('should refresh trace viewer while test is running', async ({ activate, cre
   });
 
   await testController.expandTestItems(/test.spec/);
-  selectTestItem(testController.findTestItems(/pass/)[0]);
+  const [testItem] = testController.findTestItems(/pass/);
+  selectTestItem(testItem);
 
   const testRunPromise = testController.run();
+  // wait for test to start
+  await waitForTestItemStatus(testItem, 'started');
```
pavelfeldman, line 110: sounds like the next line should take care of it since you are polling?

tests/trace-viewer.spec.ts
```diff
@@ -15,9 +15,17 @@
  */
 
 import { enableConfigs, expect, selectConfig, selectTestItem, test, traceViewerInfo } from './utils';
+import { TestItem } from './mock/vscode';
 
 test.skip(({ traceViewerMode }) => !traceViewerMode);
 
+async function waitForTestItemStatus(testItem: TestItem, status: TestItem['status']) {
```
pavelfeldman, line 22: It does not sound like we should need it, it is better to make the assertion that follows a waiting assertion.

tests/mock/vscode.ts
```diff
@@ -193,6 +193,8 @@ export class TestItem {
   tags: readonly TestTag[] = [];
   canResolveChildren = false;
   status: 'none' | 'enqueued' | 'started' | 'skipped' | 'failed' | 'passed' = 'none';
+  _didChangeStatus = new EventEmitter<TestItem>();
+  onDidChangeStatus = this._didChangeStatus.event;
```
pavelfeldman, line 197: I don't see this on vscode.TestItem,

ruifigueira: I'll use `TestController.onDidCreateTestRun > TestRun.onDidChange` to achieve the same goal.

tests/trace-viewer.spec.ts
ruifigueira: My problem is that the polling fails with timeout, at least in my computer, and that's because tests are taking more than 5 seconds to start.

That `waitForTestItemStatus(testItem, 'started')` is just to ensure test is in a state where I have confidence I won't get a timeout while polling.

ruifigueira:
Rebased it again, and adapted it to new changes.

@pavelfeldman I noticed the `embeddedTraceViewer` setting is not materialized in the actual VS Code, and I found no path to enable it except in the mocked vscode.

I had to change the value directly in the code to be able to test it in VS Code. Is there any other way?

PR #519, opened 2024-08-18
Trace Viewer as webviewView

ruifigueira:
As requested in https://github.com/microsoft/playwright-vscode/pull/513#discussion_r1714624850

It's still in draft:

- when view is disposed or it isn't resolved yet (e.g view was explicitly closed and then VS code was restarted), I couldn't find a way of programatically show the view. I need to explicitly open the view with "View: Open View" command. I even tried to execute that command programatically, but it always shows a quick pick to pick the view, I cannot pass the view ID as a parameter
- I added a `when` condition to this view's contribution point, but it only checks `showTrace` configuration because `embeddedTraceViewer` is not a declarative configuration yet
- I already separated testing harness changes into a different commit, so if you want it in a different PR it's easy to create (for the time being, I'm letting it this way to understand the context of the changes, which are mostly to support dynamic `webview.html` changes)

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,6 +122,12 @@
           "type": "webview",
           "id": "pw.extension.settingsView",
           "name": "%views.test.pw.extension.settingsView%"
+        },
+        {
+          "type": "webview",
+          "id": "pw.extension.embeddedTraceViewerView",
+          "name": "%views.test.pw.extension.embeddedTraceViewerView%",
+          "when": "config.playwright.showTrace"
         }
       ]
     }
--- a/package.nls.json
+++ b/package.nls.json
@@ -17,5 +17,6 @@
   "configuration.playwright.env": "Environment variables to pass to Playwright Test.",
   "configuration.playwright.reuseBrowser": "Show & reuse browser between tests.",
   "configuration.playwright.showTrace": "Show Trace Viewer.",
-  "views.test.pw.extension.settingsView": "Playwright"
+  "views.test.pw.extension.settingsView": "Playwright",
+  "views.test.pw.extension.embeddedTraceViewerView": "Trace Viewer"
 }
--- a/src/embeddedTraceViewer.ts
+++ b/src/embeddedTraceViewer.ts
@@ -17,27 +17,25 @@
 import type { TestConfig } from './playwrightTestTypes';
 import { getNonce } from './utils';
 import * as vscodeTypes from './vscodeTypes';
-import { DisposableBase } from './disposableBase';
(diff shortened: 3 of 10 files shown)


pavelfeldman:
We've been reviewing the trace viewer integration in VS Code with the team and so far, we have not been able to identify the scenario where it is useful. With the structure of the trace viewer as it is, it takes a lot of screen real estate and fitting it into VS Code becomes artificial. It also blurs otherwise clear context separation between the tools. As a result we did not reach a consensus of exposing it to the user.

ruifigueira:
I understand the challenges you pointed out, but I think that this could provide a better user experience.

For example, selecting actions can open its corresponding code:

https://github.com/user-attachments/assets/4d0e7d7d-1267-4b34-97ac-36b7904e4a8b

What if trace viewer UI was split into multiple views / panels / editors?

pavelfeldman:
> For example, selecting actions can open its corresponding code:

That's true, but you can see how snapshot is pretty much useless in this layout, it is too small. Btw, we can still make it reveal source in vscode, we should just hide sources tab in this mode and make reveal handle it in the ide.

> What if trace viewer UI was split into multiple views / panels / editors?

In my experience this only makes the things worse. Distributed views are harder to manage and the clear separation between windows in this case helps user separate concerns clearly as well. That is why we never integrated inspected page into chrome devtools - left had side was for page, right hand side was for tools. Easy to keep them apart in the brain.

ruifigueira:
Ok, I cannot argue against that. Nevertheless, maybe some improvements made can be adapted into the "spawn" trace viewer. 

One thing I noticed is that embedded trace viewer opens much faster the first time because it relies on model test server, which is already running.

So, instead of spawning a new process, maybe adding a way to simply open a trace viewer app from a model's test server. It's less processes running and trace viewer should be displayed faster (currently it takes a few noticeable seconds for it to start, at least in windows).

Another thing is a better integration between VS Code and trace viewer. Having a communication channel between the two would help implementing features like the code reveal on action selection described above, synchronize dark / light mode, snapshot inspecting / recording to fix problematic code...

ruifigueira:
@pavelfeldman I created two draft PRs to "materialize" my proposals from https://github.com/microsoft/playwright-vscode/pull/519#issuecomment-2342325309:

- https://github.com/microsoft/playwright-vscode/pull/531
- https://github.com/microsoft/playwright/pull/32629

As a proof of concept of a better VS Code / Trace Viewer integration, I also included the feature of action selection triggering VS code to open its corresponding code.

PR #525, opened 2024-08-30
fix: stop global setup when debugger is terminated

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32387

We're using the cancellation token passed by VS Code to abort an ongoing debug run. There's two problems with this:

1. VS Code doesn't cancel the token if the debugger is terminated
1. If the debugger terminates during `runGlobalSetup`, the underlying process will go away and stall `runGlobalSetup`.

This PR fixes both by adding the debugger termination onto the cancellation token and racing `runGlobalSetup` against it.

### DIFF
--- a/src/playwrightTestServer.ts
+++ b/src/playwrightTestServer.ts
@@ -107,14 +107,14 @@ export class PlaywrightTestServer {
       teleReceiver.dispatch(message);
   }
 
-  async runGlobalHooks(type: 'setup' | 'teardown', testListener: reporterTypes.ReporterV2): Promise<'passed' | 'failed' | 'interrupted' | 'timedout'> {
+  async runGlobalHooks(type: 'setup' | 'teardown', testListener: reporterTypes.ReporterV2, token: vscodeTypes.CancellationToken): Promise<'passed' | 'failed' | 'interrupted' | 'timedout'> {
     const testServer = await this._testServer();
     if (!testServer)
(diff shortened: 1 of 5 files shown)


src/playwrightTestServer.ts
```diff
@@ -255,11 +255,12 @@ export class PlaywrightTestServer {
       if (!locations && !testIds)
         return;
 
-      const result = await this._runGlobalHooksInServer(debugTestServer, 'setup', reporter);
+      const result = await Promise.race([
+        this._runGlobalHooksInServer(debugTestServer, 'setup', reporter),
```
dgozman, line 259: I think we should pass the cancellation token to `_runGlobalHooksInServer()` instead, and make sure to pass it from any other callsites.

dgozman:
The change looks good, but bots are unhappy. Do you think we can add a test for this?

Skn0tt:
Fixed the bots.  https://github.com/microsoft/playwright-vscode/pull/525/commits/7f02bbbd51b961a593a043621a50dada3bdcde7b adds a test that fails on `main` and passes on this branch.

Skn0tt:
Ha, turns out there was another bug inside the legacy route!

Skn0tt:
@dgozman could you take another look?

dgozman (approved):
Looks good!

PR #532, opened 2024-09-17
chore: add @babel/plugin-proposal-decorators to babelBundle

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/32619

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
     }
   },
   "scripts": {
-    "esbuild": "esbuild ./src/babelBundle.ts ./src/extension.ts ./src/oopReporter.ts ./src/debugTransform.ts  ./src/playwrightFinder.ts --bundle --outdir=out --external:vscode --external:./babelBundle --external:./debugTransform --external:./oopReporter --external:./playwrightFinder --format=cjs --platform=node --target=ES2019",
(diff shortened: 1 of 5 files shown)


src/babelBundle.ts
```diff
@@ -15,10 +15,13 @@
  */
 
 export { types as t } from '@babel/core';
-import { parse } from '@babel/parser';
+import { parse } from '@babel/core';
```
mxschmitt, line 18: Note: The plugin format between `@babel/parser` and `@babel/core` are different, we use the `@babel/core` one in order to provide similar parsing capabilities compared to upstream. We might should re-visit the plugins we add here, so they are aligned [with upstream](https://github.com/microsoft/playwright/blob/f6219e6e793054a89e89868f2277f194586086ea/packages/playwright/bundles/babel/src/babelBundleImpl.ts#L35).

package.json
```diff
@@ -165,7 +165,7 @@
     "@typescript-eslint/parser": "^7.3.1",
     "@vscode/l10n-dev": "^0.0.24",
     "@vscode/vsce": "^3.0.0",
-    "esbuild": "^0.15.15",
+    "esbuild": "^0.23.1",
```
mxschmitt, line 168: I had to update esbuild in order to make it bundle browserslist. Browserslist is a dependency by `node_modules/@babel/helper-compilation-targets`.

pavelfeldman (approved):

PR #537, opened 2024-09-27
fix: stop global setup when debugger is terminated attempt #2

Skn0tt:
Second attempt at #525, which got reverted.

The problem was that we were stopping everything as soon as the first debug session stops. It turns out that there's more than one debug session though! The child process that's used to evaluate `playwright.config.ts` gets its own child debug session. Once it successfully closes, the `onDidTerminateDebugSession` handler is executed. In the past attempt, this closed debugging.

The fix is to only call `debugEnd.cancel()` if the main debug session is closed.

Closes https://github.com/microsoft/playwright/issues/32387

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -538,7 +538,10 @@ export class Extension implements RunHooks {
       testRun.token.onCancellationRequested(() => debugEnd.cancel());
 
       let mainDebugRun: vscodeTypes.DebugSession | undefined;
-      this._vscode.debug.onDidStartDebugSession(session => { mainDebugRun ??= session; });
+      this._vscode.debug.onDidStartDebugSession(session => {
+        if (session.name === 'Playwright Test')
+          mainDebugRun ??= session;
+      });
       this._vscode.debug.onDidTerminateDebugSession(session => {
         // child processes have their own debug sessions,
         // but we only want to stop debugging if the user cancels the main session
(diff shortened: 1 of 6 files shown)


src/extension.ts
```diff
@@ -534,7 +534,19 @@ export class Extension implements RunHooks {
     };
 
     if (mode === 'debug') {
-      await model.debugTests(items, testListener, testRun.token);
+      const debugEnd = new this._vscode.CancellationTokenSource();
```
dgozman, line 537: I now think we should move this code to the place that calls `vscode.debug.startDebugging`, instead of having it here.

src/playwrightTestServer.ts
```diff
@@ -226,6 +226,23 @@ export class PlaywrightTestServer {
     let debugTestServer: TestServerConnection | undefined;
     const disposables: vscodeTypes.Disposable[] = [];
     try {
+      const debugEnd = new this._vscode.CancellationTokenSource();
+      token.onCancellationRequested(() => debugEnd.cancel());
+
+      let mainDebugRun: vscodeTypes.DebugSession | undefined;
+      this._vscode.debug.onDidStartDebugSession(session => {
+        if (session.name === 'Playwright Test')
```
mxschmitt, line 234: can't we use the `debugSessionName` constant here?
```suggestion
        if (session.name === debugSessionName)
```

dgozman (approved):

PR #538, opened 2024-10-07
fix: pass relative config path to test-server for running

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32959. Similar to the fix we shipped for debugging in #482 and #484.

### DIFF
--- a/src/playwrightTestServer.ts
+++ b/src/playwrightTestServer.ts
@@ -199,26 +199,20 @@ export class PlaywrightTestServer {
     disposable.dispose();
   }
 
-  private _configPath() {
-    // Important, VSCode will change c:\\ to C:\\ in the program argument.
-    // This forks globals into 2 worlds, hence we specify it relative than absolute.
-    return path.relative(this._model.config.workspaceFolder, this._model.config.configFile);
-  }
-
-  private _ensureUpperCaseDriveLetter(cwd: string) {
-    if (process.platform !== 'win32')
-      return cwd;
-
-    // Important, VSCode gives us paths starting with c:\\
-    // Using that as CWD leads to https://github.com/microsoft/playwright/issues/9193,
-    // which is very hard to fix.
-    // We can work around by always uppercasing our CWD drive letter.
-    return cwd[0].toUpperCase() + cwd.substring(1);
+  private _normalizePaths() {
+    let cwd = this._model.config.workspaceFolder;
+    if (process.platform === 'win32') {
+      // TODO: long comment
+      cwd = cwd[0].toUpperCase() + cwd.substring(1);
+    }
+    return {
+      cwd,
+      cli: path.relative(cwd, this._model.config.cli),
+      config: path.relative(cwd, this._model.config.configFile),
+    };
   }
 
   async debugTests(items: vscodeTypes.TestItem[], runOptions: PlaywrightTestRunOptions, reporter: reporterTypes.ReporterV2, token: vscodeTypes.CancellationToken): Promise<void> {
-    const args = ['test-server', '-c', this._configPath()];
-
     const addressPromise = new Promise<string>(f => {
       const disposable = this._options.onStdOut(output => {
         const match = output.match(/Listening on (.*)/);
@@ -234,11 +228,13 @@ export class PlaywrightTestServer {
     let debugTestServer: TestServerConnection | undefined;
     let disposable: vscodeTypes.Disposable | undefined;
     try {
+      const paths = this._normalizePaths();
+
       await this._vscode.debug.startDebugging(undefined, {
         type: 'pwa-node',
         name: debugSessionName,
         request: 'launch',
-        cwd: this._model.config.workspaceFolder,
+        cwd: paths.cwd,
         env: {
           ...process.env,
           CI: this._options.isUnderTest ? undefined : process.env.CI,
@@ -250,8 +246,8 @@ export class PlaywrightTestServer {
           PW_TEST_SOURCE_TRANSFORM_SCOPE: testDirs.join(pathSeparator),
           PWDEBUG: 'console',
         },
-        program: this._model.config.cli,
-        args,
+        program: paths.cli,
+        args: ['test-server', '-c', paths.config],
       });
 
       if (token?.isCancellationRequested)
@@ -327,15 +323,15 @@ export class PlaywrightTestServer {
   }
 
   private async _createTestServer(): Promise<TestServerConnection | null> {
-    const cwd = this._model.config.workspaceFolder;
+    const paths = this._normalizePaths();
     const args = [
-      path.relative(cwd, this._model.config.cli),
+      paths.cli,
       'test-server',
-      '-c', this._configPath()
+      '-c', paths.config,
     ];
     const wsEndpoint = await startBackend(this._vscode, {
       args,
-      cwd: this._ensureUpperCaseDriveLetter(cwd),
+      cwd: paths.cwd,
       envProvider: () => {
         return {
           ...this._options.envProvider(),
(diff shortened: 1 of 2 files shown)


Skn0tt:
hmm, doesn't work on windows in my testing. looking more.

Skn0tt:
Found the culprit! The real root cause is https://github.com/microsoft/playwright/issues/9193. It still exists, but is hard to solve on Playwright's side. We can work around it by normalising the CWD we use to spawn the test-server, and that's what https://github.com/microsoft/playwright-vscode/pull/538/commits/f311510bb3546d3092c614d90058fe329f8adc82 does.

src/playwrightTestServer.ts
```diff
@@ -199,11 +199,25 @@ export class PlaywrightTestServer {
     disposable.dispose();
   }
 
-  async debugTests(items: vscodeTypes.TestItem[], runOptions: PlaywrightTestRunOptions, reporter: reporterTypes.ReporterV2, token: vscodeTypes.CancellationToken): Promise<void> {
+  private _configPath() {
     // Important, VSCode will change c:\\ to C:\\ in the program argument.
```
Skn0tt, line 203: I think this is the wrong way around. If you look at https://github.com/microsoft/vscode-js-debug/blob/e620efe1841742402a332b532df2aa2ec95361f6/src/common/pathUtils.ts#L177, you can see that it defaults to lowercasing letters. I can't find any callsite that passes the parameter to make it use uppercase parameters. The comment in https://github.com/microsoft/vscode-js-debug/blob/e620efe1841742402a332b532df2aa2ec95361f6/src/common/pathUtils.ts#L176 also states that VS Code prefers lowercase.

Besides that, if VSCode really changed`c:\\` to `C:\\`, we wouldn't need to work around things because `C:\\` doesn't break Playwright's loader - only `c:\\` does.

```suggestion
    // Important, VSCode will change C:\\ to c:\\ in the program argument.
```

Skn0tt: Ignore the previous comment. Executing this, I found that it passes the program argument as a relative path, but ensures that CWD is uppercased:

image
image

```suggestion
    // Important, VSCode will change c:\\ to C:\\ in the CWD.
```

Skn0tt: Nope, all wrong! The existing is right, see my next comment.

Skn0tt:
I'll summarize the behaviour we're working around here for future visitors.

The Windows Filesystem is case-insensitive, but Node.js module loading is case-sensitive. That means that on Windows, `C:\foo` and `c:\foo` point to the same file, but on Node.js `require`-ing both of them will result in two instances of the file. This can lead to two instances of `@playwright/test` being loaded, which can't happen.

You can trigger this behaviour by running `npx playwright test C:\tests\foo.spec.js c:\tests\bar.spec.js` for example. Both of them import `@playwright/test`, but then one of them loads `C:\node_modules\playwright` and the other loads `c:\node_modules\playwright`. There's an existing comment in the code that describes this as `This forks globals into 2 worlds`.

Working around this is easy: Always specify file paths in the same case. https://github.com/microsoft/playwright-vscode/pull/482 did this by making the paths relative.

There's another weird behaviour that plays into this, though. If the `c:\foo.js` file requires from `node_modules`, then the Node.js resolution algorithm will sometimes call [`realpath`](https://github.com/nodejs/node/blob/e79ae1bf0c82489d80b5f3eb2bd9b089fdfafb36/lib/internal/modules/cjs/loader.js#L774). This uses some Windows-specific implementation, and it just so happens that `realpath` normalises `c:\node_modules\lib` into `C:\node_modules\lib`. It uppercases the drive letter, leading to the same buggy behaviour we've seen before.

**What does this mean?** Never ever use paths with lowercase drive letters, and don't be in a CWD that has a lowercase drive letter.

Powershell realised this and made it [impossible to go into a lowercase CWD](https://github.com/microsoft/playwright/issues/9193#issuecomment-1219362150), we did in https://github.com/microsoft/playwright-vscode/pull/482 when we swapped absolute paths with lowercase drive letters for relative ones, and VS Code realised it when they [normalize the `program` argument](https://github.com/microsoft/playwright/issues/9193#issuecomment-1219362150).

For some reason, `this._model.config.workspaceFolder`, which we get from VS Code, [has a lower-case drive letter](https://github.com/microsoft/vscode-js-debug/blob/e620efe1841742402a332b532df2aa2ec95361f6/src/common/pathUtils.ts#L176) though. When we spawn a child process on that CWD in `_startBackend`, this leads to the bug described above. And this PR fixes it by uppercasing the drive letter.

src/playwrightTestServer.ts
```diff
@@ -313,10 +327,15 @@ export class PlaywrightTestServer {
...
+    const cwd = this._model.config.workspaceFolder;
+    const args = [
+      path.relative(cwd, this._model.config.cli),
+      'test-server',
+      '-c', this._configPath()
+    ];
     const wsEndpoint = await startBackend(this._vscode, {
       args,
-      cwd: this._model.config.workspaceFolder,
+      cwd: this._ensureUpperCaseDriveLetter(cwd),
```
dgozman, line 338: - Shouldn't we do the same at lines 241 and 253 new code, inside `debugTests()`?
- Could line 360 new code that uses `require.resolve()` have the same issue?
- What about all the same places in `playwrightTestCLI.ts`?

src/playwrightTestServer.ts
```diff
@@ -313,10 +327,15 @@ export class PlaywrightTestServer {
   }
 
   private async _createTestServer(): Promise<TestServerConnection | null> {
-    const args = [this._model.config.cli, 'test-server', '-c', this._model.config.configFile];
+    const cwd = this._model.config.workspaceFolder;
+    const args = [
+      path.relative(cwd, this._model.config.cli),
```
dgozman, line 332: This one deserves the same comment as in the `_configPath()` method. I wonder whether it's better to rename `_configPath()` into `_relativePath()` and reuse? Or perhaps `_normalizePaths()` that would return `{ cwd, cli, config }` triplet with all the right relativeness/casing and have a single comment explaining the situation?

src/playwrightTestServer.ts
```diff
@@ -199,11 +199,25 @@ export class PlaywrightTestServer {
...
-    const configFile = path.relative(this._model.config.workspaceFolder, this._model.config.configFile);
-    const args = ['test-server', '-c', configFile];
+    return path.relative(this._model.config.workspaceFolder, this._model.config.configFile);
+  }
+
+  private _ensureUpperCaseDriveLetter(cwd: string) {
+    if (process.platform !== 'win32')
+      return cwd;
+
+    // Important, VSCode gives us paths starting with c:\\
```
dgozman, line 212: I'd love you to consolidate this comment and comment in the line 203 into a single one, and either copy from the great PR comment explaining things (my preference), or link to that PR comment.

src/playwrightTestServer.ts
Skn0tt: - no, because that's a call to VS Code. VS Code already performs the disk letter uppercasing for us, see the existing comment. I'll still update it for consistency though - can't hurt to normalize on our end.
- it could have the same issue, yes. but it wouldn't manifest in the same way, because `oopReporter` doesn't import any Playwright singletons.
- could have the same problems, but i'm less sure about it there because it uses `configFolder` as the CWD. Let's ship changes to that in a separate PR and release, so we can roll back more easily.

src/playwrightTestServer.ts
Skn0tt: done!

src/playwrightTestServer.ts
Skn0tt: love the triplet idea!

dgozman (approved):

Skn0tt:
I've done one final check on my Windows machine, and this still fixes the issue. I'll merge and release this PR, and extend the patch to `playwrightTestCLI.ts` if we haven't heard any bug reports in a couple weeks.

PR #545, opened 2024-10-22
fix: skip internal Playwright internals during debugging

mxschmitt:


src/playwrightTestCLI.ts
```diff
@@ -197,6 +197,7 @@ export class PlaywrightTestCLI {
         name: debugSessionName,
         request: 'launch',
         cwd: configFolder,
+        skipFiles: ['<node_internals>/**', '**/node_modules/playwright/**', '**/node_modules/playwright-core/**'],
```
mxschmitt, line 200: `'<node_internals>/**'` is the default value and we extend it here a little bit. Any other packages we should add?

pavelfeldman: Why not add all node_modules?

pavelfeldman (approved):

src/playwrightTestCLI.ts
mxschmitt: As per offline discussion:

> I was worried that users want to sometimes debug their dependencies and step into it. If we add all node_modules they won't be able to do that, I think thats not what users expect.

PR #547, opened 2024-10-24
chore: render Error.cause

mxschmitt:
This makes `Error.cause` render in test errors:

image

When looking at the Test Output tab, they are already getting shown, due to base reporter:

image

https://github.com/microsoft/playwright/issues/26848

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -742,15 +742,10 @@ export class Extension implements RunHooks {
     return testMessage;
   }
 
-  private _formatError(error: reporterTypes.TestError, tab: string = ''): string {
+  private _formatError(error: reporterTypes.TestError): string {
     const tokens = [error.stack || error.message || error.value || ''];
-    if (error.cause) {
-      tab += '  ';
-      tokens.push(indent('[cause]: ' + this._formatError(error.cause, tab), tab));
-    }
-    function indent(lines: string, tab: string) {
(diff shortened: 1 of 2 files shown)


src/upstream/reporter.d.ts
```diff
@@ -691,6 +691,13 @@ export interface Location {
  * Information about an error thrown during test execution.
  */
 export interface TestError {
+  /**
+   * Error cause. Set when there is a
+   * [cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) for the
+   * error. Will be null if there is no cause or if the cause is not an instance of [Error] (or its subclass).
+   */
+  cause?: TestInfoError;
```
yury-s, line 699: The type of this should be TestError (i.e. reporter type, not the worker's type).

src/extension.ts
```diff
@@ -742,6 +742,18 @@ export class Extension implements RunHooks {
     return testMessage;
   }
 
+  private _formatError(error: reporterTypes.TestError, tab: string = ''): string {
+    const tokens = [error.stack || error.message || error.value || ''];
+    if (error.cause) {
+      tab += '  ';
```
yury-s, line 748: tab -> indent?

yury-s (approved):
I don't quite like the ladder, maybe format it as a flat list with `[cause]` markers?

PR #548, opened 2024-10-25
feature: send enabled projects on wire for global-setup

rahul-sharma-uipath:
sends enabled projects across the wire to playwright test server

currently `params` isn't read by the listener, so this shouldn't cause any issues

below is reference code for listener, where params is unused -
https://github.com/microsoft/playwright/blob/ff5f1628dc8dbbb0217554adc3105bff4e0a504c/packages/playwright/src/runner/testServer.ts#L149


rahul-sharma-uipath:
this change appears to affect a number of tests, but I want to see if this approach will be accepted before I go down the path of fixing them all

PR #550, opened 2024-10-25
fix(legacy): extend path normalisation to legacy CLI integration

Skn0tt:
Extends https://github.com/microsoft/playwright-vscode/pull/538 to the CLI integration, as per https://github.com/microsoft/playwright-vscode/pull/538#discussion_r1794844011.


Skn0tt:
seems to be breaking lots of tests. let's keep it as is

PR #553, opened 2024-11-11
chore(watch): resolve changed files into test IDs

Skn0tt:
This PR changes how we specify which tests to run in watch mode. Before, we had one run for changes to watched files, and a separate one for changes to watched tests. One passed `locations`, one passed `testIDs`. This PR makes us resolve files into test IDs, and use a single run that only specifies `testIDs`. This moves the extension closer in line with what the UI or watch mode do.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -569,14 +569,29 @@ export class Extension implements RunHooks {
     return testListener;
   }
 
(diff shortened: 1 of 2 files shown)


Skn0tt:
let's do this once the legacy is dropped

PR #555, opened 2024-11-14
chore: delete embedded trace viewer

Skn0tt:
We decided not to ship the embedded trace viewer. I have some changes to the trace viewer that broke the tests for it, so this PR removes the code altogether.


dgozman (approved):

ruifigueira:
I was a victim of my own code:

https://github.com/microsoft/playwright/pull/33643#issuecomment-2481249329

It's now time to let it go (well, sort of, because git doesn't let, this is probably the equivalent to purgatory).

PR #557, opened 2024-11-14
chore: fix ansi escaping for diffs

pavelfeldman:
Screenshot from 2024-11-14 11-38-28

### DIFF
--- a/src/ansi2html.ts
+++ b/src/ansi2html.ts
@@ -1,0 +1,161 @@
+/*
+  Copyright (c) Microsoft Corporation.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+export function ansi2html(text: string, defaultColors?: { bg: string, fg: string }): string {
+  const regex = /(\x1b\[(\d+(;\d+)*)m)|([^\x1b]+)/g;
+  const tokens: string[] = [];
+  let match;
+  let style: any = {};
+
+  let reverse = false;
+  let fg: string | undefined = defaultColors?.fg;
+  let bg: string | undefined = defaultColors?.bg;
+
+  while ((match = regex.exec(text)) !== null) {
+    const [, , codeStr, , text] = match;
+    if (codeStr) {
+      const code = +codeStr;
+      switch (code) {
+        case 0: style = {}; break;
+        case 1: style['font-weight'] = 'bold'; break;
+        case 2: style['opacity'] = '0.8'; break;
+        case 3: style['font-style'] = 'italic'; break;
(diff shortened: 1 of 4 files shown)


src/ansi2html.ts
```diff
@@ -0,0 +1,161 @@
...
+      }
+    } else if (text) {
+      let token = escapeHTML(text);
+      const isBold = style['font-weight'] === 'bold';
+      if (isBold)
+        token = `<b>${token}</b>`;
+      const isItalic = style['font-style'] === 'italic';
+      if (isItalic)
+        token = `<i>${token}</i>`;
+      const hasOpacity = style['opacity'] === '0.8';
```
yury-s, line 115: What is so special about 0.8?

src/ansi2html.ts
```diff
@@ -0,0 +1,161 @@
...
+        case 107:
+          bg = brightAnsiColors[code - 100];
+          break;
+      }
+    } else if (text) {
+      let token = escapeHTML(text);
+      const isBold = style['font-weight'] === 'bold';
+      if (isBold)
+        token = `<b>${token}</b>`;
+      const isItalic = style['font-style'] === 'italic';
```
yury-s, line 112: Let's have a type for `style` and also just check if style.italic is true?

src/ansi2html.ts
```diff
@@ -0,0 +1,161 @@
...
+      if (isItalic)
+        token = `<i>${token}</i>`;
+      const hasOpacity = style['opacity'] === '0.8';
+      if (hasOpacity)
+        token = `<span style='color:#666;'>${token}</span>`;
+      const color = reverse ? (bg || '#000') : fg;
```
yury-s, line 118: I believe default background will depend on dark/light mode.

src/ansi2html.ts
```diff
@@ -0,0 +1,161 @@
...
+  3: '#ffcc66',
+  4: '#44a8f2',
+  5: '#b084eb',
+  6: '#afdab6',
+  7: '#fff',
+};
+
+function escapeHTML(text: string): string {
+  return text.replace(/[&"<>\s\n]/g, c => ({
+    ' ': '&nbsp;',
```
yury-s, line 154: this won't match other whitespaces, such as \t. Also, do we want to map all whitespaces to non-breaking?

src/ansi2html.ts
pavelfeldman: It is assigned in the line 34, the idea is that we use upstream code for style definition, but then we sanitize it for vscode to accept it.

src/ansi2html.ts
pavelfeldman: That would fork the code from the upstream a lot.

src/ansi2html.ts
pavelfeldman: Yeah, I checked it in both modes and it is ok because we only do reverse on mid-light background.

src/ansi2html.ts
pavelfeldman: This has not changed

src/ansi2html.ts
yury-s: This part is already different

src/ansi2html.ts
yury-s: hmm, this is what I see upstream:

```js
function escapeHTML(text: string): string {
  return text.replace(/[&"<>]/g, c => ({ '&': '&amp;', '"': '&quot;', '<': '&lt;', '>': '&gt;' }[c]!));
}
```

PR #563, opened 2024-11-23
fix: consistently enable first config

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/32338

See https://github.com/microsoft/playwright/issues/32338#issuecomment-2495555514 for more information.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -270,6 +270,8 @@ export class Extension implements RunHooks {
     this._testTree.startedLoading();
 
     const configFiles = await this._vscode.workspace.findFiles('**/*playwright*.config.{ts,js,mts,mjs}', '**/node_modules/**');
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -270,6 +270,8 @@ export class Extension implements RunHooks {
     this._testTree.startedLoading();
 
     const configFiles = await this._vscode.workspace.findFiles('**/*playwright*.config.{ts,js,mts,mjs}', '**/node_modules/**');
+    // configFiles are not sorted, sort them to ensure consistent order when we enable the first model.
```
mxschmitt, line 273:
```suggestion
    // findFiles returns results in a non-deterministic order - sort them to ensure consistent order when we enable the first model by default.
```

pavelfeldman (approved):

PR #564, opened 2024-11-26
fix: show double test output

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/30815

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -335,7 +335,7 @@ export class Extension implements RunHooks {
     if (this._testRun && !request.continuous)
       return;
 
-    await this._queueTestRun(request.include, isDebug ? 'debug' : 'run');
+    await this._queueTestRun(request, isDebug ? 'debug' : 'run');
 
     if (request.continuous) {
       for (const model of this._models.enabledModels())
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -391,54 +391,30 @@ export class Extension implements RunHooks {
     }
   }
 
-  private async _runTests(include: readonly vscodeTypes.TestItem[] | undefined, mode: 'run' | 'debug' | 'watch') {
+  private async _runTests(request: vscodeTypes.TestRunRequest, mode: 'run' | 'debug' | 'watch') {
     this._completedSteps.clear();
     this._executionLinesChanged();
 
-    // Create a test run that potentially includes all the test items.
```
mxschmitt, line 398: This can be removed since https://github.com/microsoft/vscode/issues/213182 has been landed upstream.

src/extension.ts
```diff
@@ -391,54 +391,42 @@ export class Extension implements RunHooks {
...
-      for (const rootItem of rootItems) {
-        if (!rootItem.children.size)
-          continue;
-        rootItem.children.forEach(c => {
-          if (!testItemForGlobalErrors)
-            testItemForGlobalErrors = c;
+      this._testController.items.forEach(item => {
+        item.children.forEach(item => {
+          if (item.children.size)
+            testItemForGlobalErrors = item;
```
pavelfeldman, line 408: This will make the last test a global error collection.

PR #569, opened 2024-12-13
feat: Select all projects checkbox

Skn0tt:
Works towards https://github.com/microsoft/playwright/issues/33865. Once we have a "unselect all" button, we can automatically select all projects by default.

https://github.com/user-attachments/assets/e542a4d7-f71e-442c-bb15-5ed84b73f3b4

### DIFF
--- a/src/settingsView.ts
+++ b/src/settingsView.ts
@@ -71,7 +71,6 @@ export class SettingsView extends DisposableBase implements vscodeTypes.WebviewV
 
     webviewView.webview.html = htmlForWebview(this._vscode, this._extensionUri, webviewView.webview);
     this._disposables.push(webviewView.webview.onDidReceiveMessage(data => {
-      console.log(data);
       if (data.method === 'execute') {
         this._vscode.commands.executeCommand(data.params.command);
       } else if (data.method === 'toggle') {
--- a/src/settingsView.ts
+++ b/src/settingsView.ts
@@ -268,9 +268,9 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
         </div>
       </div>
       <div class="section-header">${vscode.l10n.t('PROJECTS')}</div>
-      <button class="select-projects" id="select-all-projects" disabled>Select all</button>
-      <button class="select-projects" id="unselect-all-projects" disabled>Unselect all</button>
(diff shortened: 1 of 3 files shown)


src/settingsView.ts
```diff
@@ -316,12 +319,34 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
           div.appendChild(label);
           projectsElement.appendChild(div);
         }
+
+        const allEnabled = projects.every(p => p.enabled);
```
agg23, line 323: You set this to a variable, then never use it, and redo the same calculation in the next line.

src/settingsView.ts
```diff
@@ -316,12 +319,34 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
...
+          checkbox.dispatchEvent(new Event('change'));
+        }
+      });
+
       const vscode = acquireVsCodeApi();
       for (const input of document.querySelectorAll('input[type=checkbox]')) {
         input.addEventListener('change', event => {
-          vscode.postMessage({ method: 'toggle', params: { setting: event.target.getAttribute('setting') } });
+          const setting = event.target.getAttribute('setting');
+          if (setting)
```
agg23, line 348: What happens if there's no setting (I don't know this VSCode API)?

agg23 (approved):

src/settingsView.ts
Skn0tt: oops!

src/settingsView.ts
Skn0tt: It's standard DOM. We have some input elements in the HTML that have a setting attribute, e.g. this:

https://github.com/Skn0tt/playwright-vscode/blob/9f109ef3ce9af487b00136196fc67810cb0e1ee2/src/settingsView.ts#L276

Before this change, there were no elements without a setting. But i've added a new `input` element that doesn't have it, so we need to check this here. Makes me think we shouldn't attach the event listener in the first place, letb me change that!

Skn0tt: adding it to the css selector is even smoother I think: https://github.com/microsoft/playwright-vscode/pull/569/commits/a1e1a8258eba61aaaa48e9fbafccd84b6eedc9e6

agg23 (approved):

PR #574, opened 2024-12-16
feat: Select all projects iconbutton

Skn0tt:
Similar to https://github.com/microsoft/playwright-vscode/pull/569, but as an icon button.

Works towards https://github.com/microsoft/playwright/issues/33865 by adding a Select All / Unselect All button. Once we have that, we can start automatically selecting all projects by default.

https://github.com/user-attachments/assets/535098fc-042c-41fd-b055-f37eff1acacb


src/settingsView.ts
```diff
@@ -318,7 +331,21 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
...
+        
+        const allEnabled = projects.every(p => p.enabled);
+        selectAllButton.hidden = allEnabled;
+        unselectAllButton.hidden = !allEnabled;
       }
+      
+      function setAllProjects(checked) {
+        for (const input of projectsElement.querySelectorAll('input[type=checkbox]')) {
+          input.checked = checked;
+          input.dispatchEvent(new Event('change'));
```
pavelfeldman, line 343: This would do a lot of things many times if you have 20 projects. I'd recommend making it a bit more MVC-y:

Option 1:
You currently dispatch an event in line 343 that you listen to in line 327. You can consider short-cutting it and making setProjectEnabled receive array of values for a batch update. That way instead of 20 whole-UI updates we get 1.

Option 2:
Going even further MVC where sidebar is a View, introduce an event "setAllProjectEnabled(boolean)" that would site next to setProjectEnabled and would do the right thing, also once.

Skn0tt: Sounds good. I implemented Option 2 in https://github.com/microsoft/playwright-vscode/pull/574/commits/8f2b484d80ef0c0b72aa1d8a48ef210ab5adff14 - Option 1 would've required keeping a list of project names within the view, which felt weird.

pavelfeldman (approved):

PR #575, opened 2024-12-16
feat: on new-opened models, enable all projects

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/33865.

### DIFF
--- a/src/testModel.ts
+++ b/src/testModel.ts
@@ -109,14 +109,12 @@ export class TestModel extends DisposableBase {
       return;
     await this._listFiles();
     if (configSettings) {
-      let firstProject = true;
       for (const project of this.projects()) {
         const projectSettings = configSettings.projects.find(p => p.name === project.name);
         if (projectSettings)
           project.isEnabled = projectSettings.enabled;
-        else if (firstProject)
-          project.isEnabled = true;
-        firstProject = false;
+        else
+          project.isEnabled = false;
       }
     } else {
       for (const project of this.projects())
--- a/src/testModel.ts
+++ b/src/testModel.ts
@@ -713,11 +713,13 @@ export class TestModelCollection extends DisposableBase {
     if (model.isEnabled === enabled)
       return;
     model.isEnabled = enabled;
-    if (userGesture)
-      this._saveSettings();
     model.reset();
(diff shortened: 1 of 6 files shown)


src/testModel.ts
```diff
@@ -715,11 +713,13 @@ export class TestModelCollection extends DisposableBase {
...
     model.isEnabled = enabled;
-    if (userGesture)
-      this._saveSettings();
     model.reset();
     const configSettings = this._configSettings(model.config);
-    model._loadModelIfNeeded(configSettings).then(() => this._didUpdate.fire());
+    model._loadModelIfNeeded(configSettings).then(() => {
+      this._didUpdate.fire();
+      if (userGesture)
+        this._saveSettings();
```
Skn0tt, line 721: Before, we were saving settings immediately before reading them in `this._configSettings`. This breaks the "no settings yet" path in `_loadModelIfNeeded`, and broke tests in`track-config.spec.ts`. By moving the save settings call after the `_loadModel`, we keep the "no settings yet" path working.

pavelfeldman: How is the following case handled:

- I am like 90% of our users
- I created a playwright project a year ago, I only have one config, 'chromium' was checked for me automatically
- For the last year I was pressing "Run" button and I was getting all my tests run against Chromium
- I don't expect this to change when I update extension.

Skn0tt: Well when I wrote this I tried to change it a way that existing users don't have any changes. This is a complicated thing though, so let me try writing a test to verify this.

Skn0tt: Oh no :/ If the user never made a gesture that invokes `_saveSettings`, then their workspace settings are undefined, making them indistinguishable from users opening a new workspace. Not sure why, but I somehow assumed that we'd save settings immediately on opening. I'll have to rethink this.

Skn0tt:
Here's a PR to add that marker: https://github.com/microsoft/playwright-vscode/pull/577

After rolling that out, we'll wait for a couple of weeks for folks to update + activate that code in their Playwright workspace. Those workspaces will then be detectable as existing workspaces, allowing us to opt them out the new project defaults.

PR #579, opened 2024-12-18
chore: running tests should create workspace-settings

Skn0tt:
We store settings like activated projects in `pw.workspace-settings`. But we only write to it if it's modified, which means that users who never change anything about them don't get their settings stored. If we ever change the default settings, this means their workflows + muscle memory will be broken. Not good!

To prevent that from happening, this PR makes us `saveSettings()` before every test run. It also adds some tests to cover that.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -129,6 +129,7 @@ export class Extension implements RunHooks {
   }
 
   async onWillRunTests(config: TestConfig, debug: boolean) {
+    this._models.onWillRunTests();
     await this._reusedBrowser.onWillRunTests(config, debug);
     return {
       connectWsEndpoint: this._reusedBrowser.browserServerWSEndpoint(),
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -129,6 +129,7 @@ export class Extension implements RunHooks {
   }
 
   async onWillRunTests(config: TestConfig, debug: boolean) {
+    this._models.onWillRunTests();
```
pavelfeldman, line 132: This one is a little funny because embedder (extension) notifies model collection that a model called the embedder and notified it about the runTests event. I would call collection.saveSettings right where the model runs tests instead to save on the indirection.

Skn0tt: It is! Didn't feel like I could just remove the `private`, so I added some indirection. But it looks like `TestModel` also calls `TestModelCollection#_modelUpdated`, so they trust one another. Made the change!

pavelfeldman (approved):

PR #580, opened 2024-12-20
chore: turn `<script>` files into Typescript

Skn0tt:


pavelfeldman:
Looks great! We can also remove "allowJs": true from tsconfig now or in the follow-up.

pavelfeldman (approved):

PR #581, opened 2024-12-20
chore: highlight HTML code

Skn0tt:
Screenshot 2024-12-20 at 17 42 50

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,5 @@
+{
+	"recommendations": [
+		"bierner.lit-html"
+	]
+}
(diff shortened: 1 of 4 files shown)


.vscode/extensions.json
```diff
@@ -0,0 +1,5 @@
+{
```
pavelfeldman, line 1: Those are some giant indentations!

src/utils.ts
```diff
@@ -170,3 +170,13 @@ export function getNonce() {
     text += possible.charAt(Math.floor(Math.random() * possible.length));
   return text;
 }
+
+// this is a no-op template tag. it instructs the "bierner.lit-html" vscode extension to highlight the string as HTML.
+export function html(strings: TemplateStringsArray, ...expressions: unknown[]) {
+  let result = strings[0];
```
pavelfeldman, line 176: copilot is saying this:

```js
return strings.reduce((acc, str, i) => acc + expressions[i - 1] + str);
```

pavelfeldman (approved):

PR #584, opened 2024-12-31
fix(windows): clear diagnostics errors

mxschmitt:
**Whats happening?**

The issue is that we write errors to a Map with "Node.js like paths" so that the drive-letter is **uppercase** but we delete them out of the map by using the VSCode workspace change event, which are VSCode like paths, which have **always lowercase drive letters**.

This PR wraps when calling `set` inside `this._vscode.Uri.file(path).fsPath` which normalises and is what we already do for other places when we interact with locations from the reporter API.

https://github.com/microsoft/playwright-vscode/blob/ffcd83afd0a4efd16c38c9ae86c44332f2484956/src/testModel.ts#L395-L396

Resolves https://github.com/microsoft/playwright/issues/34146
Resolves https://github.com/microsoft/playwright/issues/33671

### DIFF
--- a/src/testModel.ts
+++ b/src/testModel.ts
@@ -201,7 +201,7 @@ export class TestModel extends DisposableBase {
     }
 
     if (report.error?.location) {
-      this._errorByFile.set(report.error?.location.file, report.error);
+      this._errorByFile.set(this._vscode.Uri.file(report.error.location.file).fsPath, report.error);
       this._collection._modelUpdated(this);
       return;
     }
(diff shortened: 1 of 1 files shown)


dgozman (commented):
Please add a test. Overall, I don't like how we sprinkle this pattern in some places, but not consistently. We need to figure out a nice boundary where we convert between Node.js paths and VSCode paths and stick to that, hopefully enforced by TypeScript. Any ideas?

dgozman:
Closing in favor of #587.

PR #588, opened 2025-01-09
test: attempt to unflake highlight-locators.spec

dgozman:


dgozman:
Unfortunately, this did not help.

PR #590, opened 2025-01-10
fix: limit decorations to prevent editor slowdown

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/34248

Updating decorations slows down the editor when the number of decorations is too high. We can work around it by limiting the number of decorations.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -687,6 +687,9 @@ export class Extension implements RunHooks {
             }
           });
         }
+
+        if (completedDecorations.length > 1000) // too many decorations slow down the editor
+          break;
       }
 
       editor.setDecorations(this._activeStepDecorationType, activeDecorations);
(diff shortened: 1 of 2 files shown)


Skn0tt:
Seems to be a known problem: https://github.com/microsoft/vscode/issues/202154

src/extension.ts
```diff
@@ -687,6 +687,9 @@ export class Extension implements RunHooks {
             }
           });
         }
+
+        if (completedDecorations.length > 1000) // too many decorations slow down the editor
```
dgozman, line 691: Perhaps a 100? I can't imagine anyone reading more than five :)

dgozman (approved):

src/extension.ts
Skn0tt: This limit is per-file, so 100 will be used up quickly. And we don't know which ones the user wants to read. So i'd stay high enough that we only prevent the performance bottleneck, and nothing else.

PR #592, opened 2025-01-10
fix: make create-playwright work under CMD

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/34266. When we create the terminal, we don't know whether it's bash, Powershell, CMD or anything else. The VS Code API also doesn't give a good API to control that. So we need our `npm init` command to work under all three shells.

https://github.com/microsoft/playwright-vscode/pull/526 fixed a bug with NPM and Powershell, but regressed CMD:

Screenshot 2025-01-10 at 13 58 12

Look at the bottom line: Under CMD, it incorrectly picks up `'--'` as the directory to create the project in.

I've played around with it, and found a command that's interpreted the same by all three shells:

*cmd*

*powershell*

*zsh*

### DIFF
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -67,7 +67,7 @@ export async function installPlaywright(vscode: vscodeTypes.VSCode) {
   if (result.includes(installDepsItem))
     args.push('--install-deps');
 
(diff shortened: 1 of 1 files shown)


dgozman (approved):
Magic!

Skn0tt:
Indeed

PR #598, opened 2025-01-17
fix: show at most one decoration per line

Skn0tt:
Follow-up to https://github.com/microsoft/playwright-vscode/pull/590.

Instead of capping the number of decorations per file, we now show at most one decoration per line. This still places a good upper limit so we don't overwhelm VS Code rendering with too many decorations. But it also prevents the confusing UI of decorations abruptly cutting off.

https://github.com/user-attachments/assets/3073f5bb-995b-436a-ac4d-de2c59c517c8

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -675,25 +675,23 @@ export class Extension implements RunHooks {
           activeDecorations.push({ range: location.range });
       }
 
-      const completedDecorations: vscodeTypes.DecorationOptions[] = [];
+      const completedDecorations: Record<number, vscodeTypes.DecorationOptions> = {};
       for (const { location, duration } of completed) {
         if (uriToPath(location.uri) === uriToPath(editor.document.uri)) {
-          completedDecorations.push({
+          const line = location.range.start.line;
+          completedDecorations[line] = {
             range: location.range,
             renderOptions: {
               after: {
                 contentText: ` \u2014 ${duration}ms`
               }
             }
-          });
+          };
         }
-
-        if (completedDecorations.length > 1000) // too many decorations slow down the editor
-          break;
       }
 
       editor.setDecorations(this._activeStepDecorationType, activeDecorations);
-      editor.setDecorations(this._completedStepDecorationType, completedDecorations);
+      editor.setDecorations(this._completedStepDecorationType, Object.values(completedDecorations));
     }
 
   }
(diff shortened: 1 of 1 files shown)


pavelfeldman (approved):
A test would be nice!

src/extension.ts
```diff
@@ -675,25 +675,23 @@ export class Extension implements RunHooks {
           activeDecorations.push({ range: location.range });
       }
 
-      const completedDecorations: vscodeTypes.DecorationOptions[] = [];
+      const completedDecorations: Record<number, vscodeTypes.DecorationOptions> = {};
       for (const { location, duration } of completed) {
         if (uriToPath(location.uri) === uriToPath(editor.document.uri)) {
```
pavelfeldman, line 680: drive-by I would pre-compute the editor's uri instead of doing it 1000 times. Maybe even store decorations by line from the get go.

Skn0tt:
> A test would be nice!

the previous PR already added a good test: https://github.com/microsoft/playwright-vscode/blob/a95208dbb61026b3369396f3afb8d0dbd530f3c7/tests/decorations.spec.ts#L90-L111

pavelfeldman (approved):

toduyemi:
Is there a way to config this? These decorations were something i relied on heavily in my debug flow, like to get a sense of how many iterations a loop took, which logic branches were seeing heavy traffic in a loop etc. This was not something that needed to be "fixed". :(

Skn0tt:
That's good feedback! This fix is mostly for performance reasons. We need to limit the decorations in some way, otherwise VSCode gets overloaded. Let's think about what a middle ground would be where there's no performance problems, but your debug flow also isn't affected.

When you debug your tests that way, do you run a single test, or do you run all tests in file? We could try only engaging this optimisation if there's more than one test being run. 
Or we could limit the decorations to 5 per line instead of 1 per line (or any other low number).

Let me know what you think would be a good solution!

toduyemi:
Gratitude for the swift reply! 

In situations where I'm paying attention to the decorations, I'm almost definitely running a single test. I generally set most loops to 10.
> We could try only engaging this optimisation if there's more than one test being run.

This seems like a great middle ground, i'd be happy with. Another potential solution would be including some sort of counter in the decoration alongside the elapsed time. So it would tell you that {the 5th execution of this line of code took 400ms} or something to that effect.

The decorators really have proven very useful while authoring a test or light debugging (means I don't have to launch debugger or PW Inspector). And since the decorations persist after the test is done, I can still refer to them as tinker with the app or test script. So i really appreciate taking this into consideration :) also love the other updates in this release. Particularly the playwright drawer!

toduyemi:
Will do! Gratitude :)

PR #601, opened 2025-01-27
chore: support fall back to config values for update snapshots

pavelfeldman:


src/settingsModel.ts
```diff
@@ -140,17 +140,17 @@ class PersistentSetting<T> extends SettingBase<T> {
...
         this.set(!this.get() as T);
       }),
     ];
   }
 
-  get(): T {
+  get(): T | undefined {
     const configuration = this._vscode.workspace.getConfiguration('playwright');
-    return configuration.get(this.settingName) as T;
+    return configuration.get(this.settingName) as T | undefined;
```
yury-s, line 153:
```suggestion
    return configuration.get(this.settingName) as (T | undefined);
```
?

yury-s (approved):

src/settingsModel.ts
pavelfeldman: should be the same thing?

PR #604, opened 2025-01-29
chore: show "one worker" next to show browser

pavelfeldman:
image


yury-s (approved):
Let's add a tooltip providing more details. One worker may be confusing for people new to playwright.

mxschmitt:
Some more ideas:
- disable parallelism 
- single worker
- 1 worker

PR #609, opened 2025-02-07
feat: decorate line with number of executions

Skn0tt:
As per feedback over here: https://github.com/microsoft/playwright-vscode/pull/598#issuecomment-2638210465

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -705,12 +705,12 @@ export class Extension implements RunHooks {
         if (uriToPath(location.uri) === editorPath) {
           const line = location.range.start.line;
           decorationCount[line] ??= 0;
-          decorationCount[line]++;
+          const count = decorationCount[line]++;
           completedDecorations[line] = {
             range: location.range,
             renderOptions: {
               after: {
-                contentText: ` \u2014 ${duration}ms${decorationCount[line] > 1 ? ` (ran ${decorationCount[line]}×)` : ''}`,
+                contentText: ` \u2014 ${duration}ms${count > 1 ? ` (ran ${count}×)` : ''}`,
               }
             }
           };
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -699,15 +699,18 @@ export class Extension implements RunHooks {
...
       for (const { location, duration } of completed) {
         if (uriToPath(location.uri) === editorPath) {
           const line = location.range.start.line;
+          const count = decorationCount[line] ?? 1;
+          decorationCount[line] = count + 1;
           completedDecorations[line] = {
             range: location.range,
             renderOptions: {
               after: {
-                contentText: ` \u2014 ${duration}ms`
```
mxschmitt, line 710: I remember we use a special character here: https://github.com/microsoft/playwright/blob/fd24521f2e11d0672852a7fe867bc3d01b8bcf94/packages/playwright-core/src/utils/stackTrace.ts#L148C41-L148C44

maybe we want to use it here as well?

mxschmitt (approved):
Voting for the special character if you like it as well!

src/extension.ts
```diff
@@ -699,15 +699,18 @@ export class Extension implements RunHooks {
           activeDecorations.push({ range: location.range });
       }
 
+      const decorationCount: Record<number, number> = {};
       const completedDecorations: Record<number, vscodeTypes.DecorationOptions> = {};
       for (const { location, duration } of completed) {
         if (uriToPath(location.uri) === editorPath) {
           const line = location.range.start.line;
+          const count = decorationCount[line] ?? 1;
+          decorationCount[line] = count + 1;
```
mxschmitt, line 708:
```suggestion
          decorationCount[line] ??= 0;
          decorationCount[line]++;
```
I think easier to understand and the same?

PR #610, opened 2025-02-11
feat: provide page snapshot as context to copilot

Skn0tt:
Enables `pageSnapshot: 'only-on-failure'` for all test runs through the extension, and injects the page snapshot into the error message that Copilot uses to fix test failures.

Screenshot 2025-02-11 at 11 00 45

related to https://github.com/microsoft/playwright/issues/34476

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -603,7 +603,7 @@ export class Extension implements RunHooks {
     const snapshot = result.attachments.find(a => a.name === 'pageSnapshot');
     if (snapshot && snapshot.path) {
       const contents = await this._vscode.workspace.fs.readFile(this._vscode.Uri.file(snapshot.path));
-      aiContext = `### Page Snapshot at Failure\n${contents.toString()}`; // cannot use ``` codeblocks, vscode markdown does not support it
+      aiContext = `### Page Snapshot at Failure\n\n${contents.toString()}`; // cannot use ``` codeblocks, vscode markdown does not support it
     }
 
     testRun.failed(testItem, result.errors.map(error => this._testMessageForTestError(error, aiContext)), result.duration);
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -520,7 +520,7 @@ export class Extension implements RunHooks {
         }
       },
 
-      onTestEnd: async (test: reporterTypes.TestCase, result: reporterTypes.TestResult) => {
+      onTestEnd: (test: reporterTypes.TestCase, result: reporterTypes.TestResult) => {
         if (result.errors.find(e => e.message?.includes(`Error: browserType.launch: Executable doesn't exist`)))
           browserDoesNotExist = true;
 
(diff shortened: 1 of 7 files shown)


src/testModel.ts
```diff
@@ -551,6 +551,7 @@ export class TestModel extends DisposableBase {
       connectWsEndpoint: showBrowser ? externalOptions.connectWsEndpoint : undefined,
       updateSnapshots: noOverrideToUndefined(this._embedder.settingsModel.updateSnapshots.get()),
       updateSourceMethod: noOverrideToUndefined(this._embedder.settingsModel.updateSourceMethod.get()),
+      pageSnapshot: 'only-on-failure',
```
Skn0tt, line 554: if we release this is the default in `@playwright/test`, we can remove this override again.

src/extension.ts
```diff
@@ -520,7 +520,7 @@ export class Extension implements RunHooks {
         }
       },
 
-      onTestEnd: (test: reporterTypes.TestCase, result: reporterTypes.TestResult) => {
+      onTestEnd: async (test: reporterTypes.TestCase, result: reporterTypes.TestResult) => {
```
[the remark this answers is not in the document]
Skn0tt: how do you like https://github.com/microsoft/playwright-vscode/pull/610/commits/2dba9b5e942902eb127cabc7b517b76c23117ff2? Or should we make it an asynchronous inline function so the code stays colocated?

src/extension.ts
```diff
@@ -597,6 +598,17 @@ export class Extension implements RunHooks {
       await installBrowsers(this._vscode, model);
   }
 
+  private async _markTestRunFailed(testRun: vscodeTypes.TestRun, testItem: vscodeTypes.TestItem, result: reporterTypes.TestResult) {
+    let aiContext: string | undefined;
+    const snapshot = result.attachments.find(a => a.name === 'pageSnapshot');
+    if (snapshot && snapshot.path) {
+      const contents = await this._vscode.workspace.fs.readFile(this._vscode.Uri.file(snapshot.path));
+      aiContext = `### Page Snapshot at Failure\n${contents.toString()}`; // cannot use ``` codeblocks, vscode markdown does not support it
```
dgozman, line 606: Don't you need two `\n` to make a paragraph?

src/extension.ts
```diff
@@ -597,6 +598,17 @@ export class Extension implements RunHooks {
       await installBrowsers(this._vscode, model);
   }
 
+  private async _markTestRunFailed(testRun: vscodeTypes.TestRun, testItem: vscodeTypes.TestItem, result: reporterTypes.TestResult) {
+    let aiContext: string | undefined;
+    const snapshot = result.attachments.find(a => a.name === 'pageSnapshot');
+    if (snapshot && snapshot.path) {
+      const contents = await this._vscode.workspace.fs.readFile(this._vscode.Uri.file(snapshot.path));
```
dgozman, line 605: I wonder whether we should either limit the number of files we read here, or make `pageSnapshot` attachment content-based, not file-based.

dgozman (approved):

src/extension.ts
Skn0tt: some renderers do, you're right. vscode didn't require it for some reason, but it still makes sense to make it two newlines in case vscode markdown rendering ever changes.

src/extension.ts
Skn0tt: It only ever reads one file per test, which should be fine.

If we make this content-based, there's a couple downsides:

- it'll auto-print to terminal on some reporters, which is very noisy
- it'll take up more memory for any consumer that holds on to old snapshots

Also, AI tools like Copilot won't be able to discover it from the file system. not yet sure if this a good or bad thing.

PR #613, opened 2025-02-18
fix: running folder should use folder path as arg

Skn0tt:
Part of https://github.com/microsoft/playwright/issues/34813. When running a folder item, we should send `folder/` to playwright, not `folder` (note the trailing slash).


Skn0tt:
Closing as per https://github.com/microsoft/playwright/pull/34838#issuecomment-2665511557

PR #614, opened 2025-02-18
fix: running a folder shouldn't also run folders that share prefix

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/34813. When running the `foo` folder, we were sending `foo` over the wire instead of `foo/`. That meant that the `foobar/` folder also gets executed, which isn't expected.

The fix is to include the trailing slash in the locations we send to Playwright. Because Playwright interprets `/root/project/dir/` as a regular expression, I made it omit the leading slash for absolute directories. That has the same effect, because Playwright's location checker isn't strict.

### DIFF
--- a/src/playwrightTestCLI.ts
+++ b/src/playwrightTestCLI.ts
@@ -18,12 +18,13 @@ import { spawn } from 'child_process';
 import path from 'path';
 import { ConfigFindRelatedTestFilesReport, ConfigListFilesReport } from './listTests';
 import { ReporterServer } from './reporterServer';
-import { escapeRegex, findNode, pathSeparator, runNode, uriToPath } from './utils';
+import { escapeRegex, findNode, pathSeparator, preventRegexLookalike, runNode, uriToPath } from './utils';
 import * as vscodeTypes from './vscodeTypes';
 import * as reporterTypes from './upstream/reporter';
 import type { PlaywrightTestOptions, PlaywrightTestRunOptions } from './playwrightTestTypes';
 import { debugSessionName } from './debugSessionName';
 import type { TestModel } from './testModel';
+import { upstreamTreeItem } from './testTree';
 
 export class PlaywrightTestCLI {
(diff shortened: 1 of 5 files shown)


src/playwrightTestServer.ts
```diff
@@ -91,7 +91,7 @@ export class PlaywrightTestServer {
     if (!connection)
       return;
     // Locations are regular expressions.
-    locations = locations.map(escapeRegex);
+    locations = locations.map(preventRegexLookalike).map(escapeRegex);
```
dgozman, line 94: Since we call `escapeRegex` - perhaps we don't need `preventRegexLookalike` here at all?

src/playwrightTestCLI.ts
```diff
@@ -118,7 +119,7 @@ export class PlaywrightTestCLI {
       'test',
       '-c', configFile,
       ...extraArgs,
-      ...escapedLocations,
+      ...escapedLocations.map(preventRegexLookalike),
```
dgozman, line 122: Since we do call `escapeRegex` already, why do we need this one?

src/playwrightTestServer.ts
Skn0tt: `escapeRegex` doesn't do anything to forward slashes. I can merge the `preventRegexLookalike` logic into `escapeRegex` if you'd prefer that.

dgozman: Yep, that would be great. Just escape the forward slashes together with other characters.

Skn0tt: Escaping the forward slashes wasn't enough, but removing the leading one did the trick: https://github.com/microsoft/playwright-vscode/pull/614/commits/42995e8ae458ffd23f11aa5b69fc1637d2d60f74

I also made sure that the exec log always shows the right trailing slash. `path.relative` likes to remove that.

dgozman (approved):

PR #619, opened 2025-03-26
Fix pnp for nested configs

goloveychuk:
fixes 
https://github.com/microsoft/playwright/issues/35370
Personally validated that it works for pnp for my setup.
But not sure if it can break other setups.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -290,7 +290,7 @@ export class Extension implements RunHooks {
 
       let playwrightInfo = null;
       try {
-        playwrightInfo = await getPlaywrightInfo(this._vscode, workspaceFolderPath, configFilePath, this._envProvider({workspaceFolder: workspaceFolderPath}));
+        playwrightInfo = await getPlaywrightInfo(this._vscode, workspaceFolderPath, configFilePath, this._envProvider({ workspaceFolder: workspaceFolderPath }));
       } catch (error) {
         if (userGesture) {
           this._vscode.window.showWarningMessage(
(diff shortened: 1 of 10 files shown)


src/playwrightTestServer.ts
```diff
@@ -201,7 +201,7 @@ export class PlaywrightTestServer {
   }
 
   private _normalizePaths() {
-    let cwd = this._model.config.workspaceFolder;
+    let cwd = path.dirname(this._model.config.configFile);
```
Skn0tt, line 204: Posting my findings here: `workspaceFolder` points to the root of the VS Code workspace, whereas `configFile` points to `playwright.config.ts`. This means that this change only makes a difference for monorepos, where `playwright.config.ts` is not located in the repo root (as you outlined in your issue).

Skn0tt (commented):
The change looks good generally. Could you try adding a test for this? If a test is hard, please add a comment to your change that explains why this is important to support PnP.

goloveychuk:
@Skn0tt I started to write integration test, with two project configs in different depth inside monorepo.
And found out that one settings.json cannot configure correct relative paths for two configs in different depth.

I propose to replace current fix with different
Add support of `${workspaceFolder}` variable inside config and replace it manually in _envProvider. Wdyt?

Skn0tt:
> Add support of ${workspaceFolder} variable inside config and replace it manually in _envProvider. Wdyt?

Could you elaborate? Do you want to add a template variable to `playwright.config.ts`?

goloveychuk:
@Skn0tt I meant playwright.env config in settings.json
https://github.com/microsoft/playwright-vscode/pull/619/commits/03adbef21b80a9f19742f1e62f9c49b819d6756d#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626R341

There is a e2e test for pnp. It creates multiconfig monorepo and runs tests for two playwright configs.
There is a bug in extensions not related to pnp though, which is caused by duplicate test file names.
Other strange thing, not sure if it's expected or not, not all configs/models are enabled by default, only first.
Which could be not expected for users.
But interestingly, I couldn't enable all models by this https://github.com/microsoft/playwright-vscode/pull/619/commits/0073d4d0f7910544c391e60a7e187e620761e4b5


Alternative to using variables in playwright.env - first class pnp support: detect .pnp.cjs file, add it to NODE_OPTIONS env var.

src/extension.ts
```diff
@@ -327,12 +327,22 @@ export class Extension implements RunHooks {
     this._workspaceObserver.setWatchFolders(this._models.testDirs());
   }
 
-  private _envProvider(): NodeJS.ProcessEnv {
+  private _envProvider: EnvProvider = ({ workspaceFolder }) => {
+    this._vscode.workspace.getWorkspaceFolder(this._context.extensionUri);
```
goloveychuk, line 331: todo rm.
I can also get workpaceFolder from `this._vscode.workspace.rootPath` but not sure if it's correct for multiworkspace editors

Skn0tt (changes requested):
Hold on, this PR is becoming very big. I'm not comfortable reviewing it in one go. Let's take a step back and discuss what you'd like to change in your issue.

goloveychuk:
Historically yarn pnp support was done by introducing `playwright.env` vscode configuration setting.
Users need to add --require and --loader node args, with relative pathes.
Now, giving that one monorepo could have multiple playwright configs in different fs depth levels, it's not possible to provide relative path to monorepo root (which has .pnp.cjs file).
I see only two solutions.
1) first class pnp support. (detect pnp.cjs in monorepo root, add it to NODE_OPTIONS)
2) introduce variables into `playwright.env` config.

My pr implements second, since it does not change previous approach/design.
Other part of PR - e2e test, and infra for this test.

Skn0tt:
> Now, giving that one monorepo could have multiple playwright configs in different fs depth levels, it's not possible to provide relative path to monorepo root (which has .pnp.cjs file).

Try providing an absolute path instead.

I'll go ahead and close this PR. Please open an issue prior to posting a PR next time, and discuss with us so we can align on a solution that we're willing to accept.

PR #624, opened 2025-04-09
chore: pick right launch options for codegen

Skn0tt:
Resolves https://github.com/microsoft/playwright/issues/26777


Skn0tt:
let me write a test first!

Skn0tt:
@dgozman please take a look.

src/reusedBrowser.ts
```diff
@@ -298,17 +300,21 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
...
     if (recordNew) {
-      await this._backend?.resetForReuse();
-      await this._backend?.navigate({ url: 'about:blank' });
+      const file = await this._createFileForNewTest(model);
+      if (!file)
+        return;
+      await model.handleWorkspaceChange({ created: new Set([file]), changed: new Set(), deleted: new Set() });
+      await model.ensureTests([file]);
+      const project = model.enabledProjects()[0];
+      if (!project)
```
dgozman, line 315: Let's fallback to the old behavior in this case.

src/extension.ts
```diff
@@ -648,6 +648,28 @@ export class Extension implements RunHooks {
...
+    if (fileItem.children.size !== 1)
+      throw new Error('expected single test in file');
+
+    const testItems = this._testTree.collectTestsInside(fileItem);
+    const testForProject = testItems.length === 1 ? testItems[0] : testItems.find(t => t.label === project.name);
+    if (!testForProject)
+      throw new Error(`Test item not found for ${project.name}`);
+
+    const request = new this._vscode.TestRunRequest([testForProject], undefined, undefined, false, true);
+    (request as any)[kTestRunOptions] = {
```
dgozman, line 664: It seems like you need to check `showBrowser` setting instead, and uncheck it after the recording has finished?

src/extension.ts
```diff
@@ -648,6 +648,28 @@ export class Extension implements RunHooks {
       await this._queueWatchRun(new this._vscode.TestRunRequest(testItems), 'items');
   }
 
+  private async _showBrowserForRecording(file: string, project: TestProject, connectWsEndpoint: string) {
+    const fileItem = this._testTree.testItemForFile(file);
+    if (!fileItem)
+      throw new Error(`File item not found for ${file}`);
```
dgozman, line 654: What happens when this throws? Does the user see an error? I think it's easy to trigger with non-standard `testMatch`. Let's be careful.

src/extension.ts
```diff
@@ -103,7 +103,7 @@ export class Extension implements RunHooks {
     });
 
     this._settingsModel = new SettingsModel(vscode, context);
-    this._reusedBrowser = new ReusedBrowser(this._vscode, this._settingsModel, this._envProvider.bind(this));
+    this._reusedBrowser = new ReusedBrowser(this._vscode, this._settingsModel, this._envProvider.bind(this), this._showBrowserForRecording.bind(this));
```
dgozman, line 106: Instead of passing this as a callback, can we create a file and run the test before calling `ReusedBrowser.record()`? This will make dependencies flow forward instead of both ways.

src/extension.ts
Skn0tt: good catch. I switched it to returns.

src/reusedBrowser.ts
```diff
@@ -345,10 +336,7 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
     });
   }
 
-  private async _createFileForNewTest(model: TestModel) {
-    const project = model.enabledProjects()[0];
-    if (!project)
-      return;
+  async createFileForNewTest(model: TestModel, project: TestProject) {
```
dgozman, line 339: I think this function does not belong here anymore and should be moved to `extension.ts`. It has nothing to do with reused browser. Can be done in a follow up.

tests/codegen.spec.ts
```diff
@@ -17,17 +17,33 @@
...
+          },
+        },
+      ]
+    }`,
   });
 
   const webView = vscode.webViews.get('pw.extension.settingsView')!;
   await webView.getByText('Record new').click();
-  await expect.poll(() => vscode.lastWithProgressData).toEqual({ message: 'recording\u2026' });
+  await webView.getByRole('checkbox', { name: 'default' }).setChecked(false);
```
dgozman, line 39: Shouldn't you update projects **before** clicking "Record new"?

dgozman (approved):

PR #628, opened 2025-04-17
fix: show config loading error

Skn0tt:
Resolves https://github.com/microsoft/playwright/issues/35544

https://github.com/user-attachments/assets/b0b6ac89-9c26-4323-96f2-dd38f9254abc

### DIFF
--- a/src/settingsView.script.ts
+++ b/src/settingsView.script.ts
@@ -23,7 +23,7 @@ const selectAllButton = document.getElementById('selectAll') as HTMLAnchorElemen
 const unselectAllButton = document.getElementById('unselectAll') as HTMLAnchorElement;
 const toggleModels = document.getElementById('toggleModels') as HTMLAnchorElement;
 
-const basename = (path: string) => path.split(/\\\//).pop()!;
+const basename = (path: string) => path.split(/[\\\/]/).pop()!;
 
 function updateProjects(projects: ProjectEntry[]) {
   const projectsElement = document.getElementById('projects') as HTMLElement;
--- a/src/settingsView.script.ts
+++ b/src/settingsView.script.ts
@@ -15,6 +15,7 @@
  */
 
 import { Config, createAction, ProjectEntry, vscode } from './common';
+import type { ConfigEntry, ModelsMessage } from './settingsView';
 
 let selectConfig: Config;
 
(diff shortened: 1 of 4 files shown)


dgozman (commented):
Looks kinda ugly. Can we show something like `Unable to load [playwright.config.ts](link to the error location)` instead?

mxschmitt (commented):
Instead of creating a new UI for showing config loading errors I'd instead try to redirect to the problem tab, this is what VSCode uses as their native experience. Maybe we can even focus to the actual Problem which we added or focus to the location of where the issue occurred and then VSCode will show it as red. I think marking it as red in the WebView is a good idea but I'd resist from showing the error there.

Skn0tt:
I couldn't find a way of programmatically opening the Problems tab on a specific problem, so implemented the "Unable to load" UI:

Screenshot 2025-04-22 at 09 01 00

src/settingsView.script.ts
```diff
@@ -97,10 +98,31 @@ window.addEventListener('message', event => {
         actionsElement.appendChild(actionElement);
     }
   } else if (method === 'models') {
-    const { configs, showModelSelector } = params;
+    const { configs, showModelSelector } = params as ModelsMessage;
     const select = document.getElementById('models') as HTMLSelectElement;
+
+    function updateConfigErrors(config: ConfigEntry) {
+      const configErrors = document.getElementById('model-errors') as HTMLParagraphElement;
```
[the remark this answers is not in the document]
dgozman: Copilot is right!

src/settingsView.script.ts
```diff
@@ -97,10 +98,34 @@ window.addEventListener('message', event => {
...
+      configErrors.textContent = '';
+      for (const error of config.errors) {
+        if (!error.message)
```
dgozman, line 108: How could this happen? Even so, why skip the link?

src/settingsView.script.ts
```diff
@@ -97,10 +98,34 @@ window.addEventListener('message', event => {
...
+        const errorParagraph = document.createElement('p');
+        errorParagraph.textContent = 'Unable to load ';
+        if (error.location) {
+          const errorLink = document.createElement('a');
+          errorLink.textContent = `${error.location.file.split('/').pop()}`;
+          errorLink.role = 'link';
+          errorLink.onclick = () => vscode.postMessage({ method: 'openFile', params: { location: error.location! } });
+          errorParagraph.appendChild(errorLink);
+        } else {
+          errorParagraph.textContent += ' config file';
```
dgozman, line 120: Let's link to the config file in this case. Otherwise, it's not really actionable for the user.

src/settingsView.script.ts
```diff
@@ -97,10 +98,34 @@ window.addEventListener('message', event => {
...
+
+    function updateConfigErrors(config: ConfigEntry) {
+      const configErrors = document.getElementById('model-errors') as HTMLParagraphElement;
+      configErrors.textContent = '';
+      for (const error of config.errors) {
+        if (!error.message)
+          continue;
+
+        const errorParagraph = document.createElement('p');
+        errorParagraph.textContent = 'Unable to load ';
```
dgozman, line 112: Since the error location might be in a different file, what about the following layout:

```
Config loading errors:
- <foo.ts:17>
- <bar.ts:42>
```

src/settingsView.ts
```diff
@@ -85,6 +92,10 @@ export class SettingsView extends DisposableBase implements vscodeTypes.WebviewV
         this._models.setAllProjectsEnabled(configFile, enabled);
       } else if (data.method === 'selectModel') {
         this._models.selectModel(data.params.configFile);
+      } else if (data.method === 'openFile') {
+        const location = data.params.location as Location;
+        const document = await this._vscode.workspace.openTextDocument(location.file);
```
dgozman, line 97: Let's honor `location.line` and `location.column`.

tests/track-configs.spec.ts
```diff
@@ -120,3 +120,32 @@ test('should remove model for config', async ({ activate }) => {
...
+  await enableConfigs(vscode, ['playwright1.config.js', 'playwright2.config.js']);
+
+  const webView = vscode.webViews.get('pw.extension.settingsView')!;
+  await expect(webView.locator('body')).toMatchAriaSnapshot(`
+    - text: CONFIGS
+    - combobox "Select Playwright Config":
+      - option "playwright1.config.js" [selected]
+      - option "playwright2.config.js"
+    - paragraph:
+      - text: "Error: kaboom"
```
dgozman, line 143: I think this needs to be rebased.

src/settingsView.script.ts
Skn0tt: like this?

Screenshot 2025-04-22 at 14 03 16

dgozman (approved):

Skn0tt:
had to fix a test 😩

dgozman (approved):

PR #631, opened 2025-04-24
chore: enable no-floating-promises ESLint rule

mxschmitt:
As usual, a lot of false-positives but also a few interesting ones.


src/testModel.ts
```diff
@@ -62,7 +62,7 @@ export class TestModel extends DisposableBase {
   private _vscode: vscodeTypes.VSCode;
   readonly config: TestConfig;
   private _projects = new Map<string, TestProject>();
-  private _playwrightTest: PlaywrightTestCLI | PlaywrightTestServer;
+  private _playwrightTest: PlaywrightTestServer | PlaywrightTestCLI;
```
mxschmitt, line 65: This improves the DX, when we command click on `_playwrightTest.Foo` it now lands in the TestServer implementation rather than in the CLI impl.

src/extension.ts
```diff
@@ -316,13 +316,13 @@ export class Extension implements RunHooks {
       await this._models.createModel(workspaceFolderPath, uriToPath(configFileUri), playwrightInfo);
     }
 
-    await this._models.ensureHasEnabledModels();
+    this._models.ensureHasEnabledModels();
```
mxschmitt, line 319: The methods inside were sync - hence changed the method signature to sync as well.

src/extension.ts
```diff
@@ -357,7 +357,7 @@ export class Extension implements RunHooks {
 
     if (request.continuous) {
       for (const model of this._models.enabledModels())
-        model.addToWatch(request.include, cancellationToken!);
+        await model.addToWatch(request.include, cancellationToken!);
```
mxschmitt, line 360: I don't see a reason to not do that. Looking at the TestServer code it means we wait now for the Watcher to ack which seems reasonable.

src/oopReporter.ts
```diff
@@ -47,7 +47,7 @@ class TeleReporter extends TeleReporterEmitter {
   }
 
   async onEnd(result: FullResult) {
-    super.onEnd(result);
+    await super.onEnd(result);
```
mxschmitt, line 50: Didn't look intentional.

src/playwrightTestServer.ts
```diff
@@ -80,7 +80,7 @@ export class PlaywrightTestServer {
       resolvePath,
     });
     for (const message of report)
-      teleReceiver.dispatch(message);
+      await teleReceiver.dispatch(message);
```
mxschmitt, line 83: Not sure about this and the following 3 places. This means that when we receive events we'll now await onEnd and onExit. Do you see any disadvantage in it or would you be careful and `void` it as well?

package.json
```diff
@@ -211,6 +211,7 @@
     "@typescript-eslint/eslint-plugin": "^7.3.1",
     "@typescript-eslint/parser": "^7.3.1",
     "@vscode/l10n-dev": "^0.0.24",
+    "@vscode/test-electron": "^2.3.8",
```
mxschmitt, line 214: Moved it now to the main package.json since we lint this folder now. Which means we need test-electron types. I checked if there is any post-install or install step and there isn't, so I don't see any reason to not put it into devDependencies.

src/extension.ts
```diff
@@ -157,17 +157,17 @@ export class Extension implements RunHooks {
...
-        this._updateVisibleEditorItems();
+        void this._updateVisibleEditorItems();
       }),
       vscode.commands.registerCommand('pw.extension.install', async () => {
         await installPlaywright(this._vscode);
       }),
       vscode.commands.registerCommand('pw.extension.installBrowsers', async () => {
         if (!this._models.hasEnabledModels()) {
-          vscode.window.showWarningMessage(messageNoPlaywrightTestsFound);
+          void vscode.window.showWarningMessage(messageNoPlaywrightTestsFound);
```
dgozman, line 170: Perhaps await instead of void in these methods?

src/extension.ts
```diff
@@ -316,13 +316,13 @@ export class Extension implements RunHooks {
...
 
-    await this._models.ensureHasEnabledModels();
+    this._models.ensureHasEnabledModels();
     this._testTree.finishedLoading();
     return configFiles;
   }
 
   private _modelsUpdated() {
-    this._updateVisibleEditorItems();
+    void this._updateVisibleEditorItems();
```
dgozman, line 325: This one seems strange - calling async method before calling a few sync ones is usually a mistake. Perhaps we can change the order of the calls? Or make this whole method async?

src/playwrightTestServer.ts
dgozman: Upstream we void instead of await, so I guess we should do the same here.

src/playwrightTestServer.ts
```diff
@@ -402,7 +402,7 @@ export class PlaywrightTestServer {
         testServer.onReport(message => {
           if (token.isCancellationRequested && message.method !== 'onEnd')
             return;
-          teleReceiver.dispatch(message);
+          void teleReceiver.dispatch(message);
```
dgozman, line 405: I'd say this should `await`, so that we resolve after `onEnd` finishes.

src/settingsModel.ts
```diff
@@ -145,7 +145,7 @@ class PersistentSetting<T> extends SettingBase<T> {
           this._onChange.fire(this.get()!);
       }),
       vscode.commands.registerCommand(`pw.extension.toggle.${settingName}`, async () => {
-        this.set(!this.get() as T);
+        void this.set(!this.get() as T);
```
dgozman, line 148: This can be `await`.

src/playwrightTestServer.ts
```diff
@@ -130,15 +130,15 @@ export class PlaywrightTestServer {
           new Promise<{ status: 'interrupted', report: [] }>(f => token.onCancellationRequested(() => f({ status: 'interrupted', report: [] }))),
         ]);
         for (const message of report)
-          teleReceiver.dispatch(message);
+          await teleReceiver.dispatch(message);
```
dgozman, line 133:
```suggestion
          void teleReceiver.dispatch(message);
```

src/playwrightTestServer.ts
```diff
@@ -130,15 +130,15 @@ export class PlaywrightTestServer {
...
         return status;
       }
       const { report, status } = await Promise.race([
         testServer.runGlobalTeardown({}),
         new Promise<{ status: 'interrupted', report: [] }>(f => token.onCancellationRequested(() => f({ status: 'interrupted', report: [] }))),
       ]);
       for (const message of report)
-        teleReceiver.dispatch(message);
+        await teleReceiver.dispatch(message);
```
dgozman, line 141:
```suggestion
        void teleReceiver.dispatch(message);
```

dgozman (approved):

dgozman (approved):

PR #634, opened 2025-05-05
chore: rename Backend to DebugController

Skn0tt:


dgozman (commented):
What's the point of this change? I'd say it should also rename `BackendServer`, remove the template and `BackendClient` inheritance from it, and rename all local variables to `_debugController` or similar. This single rename in isolation just brings more segmentation to the codebase.

Skn0tt:
The point was that I got confused reading this code, and felt like this single change made it less confusing to me. But I agree, a bigger refactor would help more.

PR #644, opened 2025-06-16
fix: move testIdAttributeName to config.use

mxschmitt:
This fixes two things:

a) `Pick Locator` command was launched before a `ReusableBrowser` was launched via a test. This now uses the first project's `'testIdAttribute'` for it.
b) 'Record at cursor' was not wired up correctly for `'testIdAttribute'`. This now uses also the project's use `'testIdAttribute'` instead of the one from the config.

Both were using a bad `'testIdAttribute'`:  Before when you had e.g. a `testIdAttribute` configured on `project1`, it was overridden again by `project2`.

Fixes https://github.com/microsoft/playwright/issues/35881

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -203,7 +203,7 @@ export class Extension implements RunHooks {
         try {
           await this._settingsModel.showBrowser.set(true);
           await this._showBrowserForRecording(file, project);
-          await this._reusedBrowser.record(model);
+          await this._reusedBrowser.record(project);
         } finally {
           await this._settingsModel.showBrowser.set(showBrowser);
         }
@@ -213,7 +213,11 @@ export class Extension implements RunHooks {
         if (!model)
           return vscode.window.showWarningMessage(messageNoPlaywrightTestsFound);
 
-        await this._reusedBrowser.record(model);
+        const project = model.enabledProjects()[0];
+        if (!project)
+          return vscode.window.showWarningMessage(this._vscode.l10n.t(`Project is disabled in the Playwright sidebar.`));
+
+        await this._reusedBrowser.record(project);
       }),
       vscode.commands.registerCommand('pw.extension.command.toggleModels', async () => {
         this._settingsView.toggleModels();
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -198,26 +198,26 @@ export class Extension implements RunHooks {
         if (!file)
           return;
 
-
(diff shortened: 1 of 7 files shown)


src/reusedBrowser.ts
```diff
@@ -242,8 +246,8 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
     return !this._isRunningTests && !!this._pageCount;
   }
 
-  async record(model: TestModel) {
-    if (!model || !this._checkVersion(model.config))
+  async record(project: TestProject) {
+    if (!project.model || !this._checkVersion(project.model.config))
```
dgozman: How could there be no `project.model`?

src/extension.ts
```diff
@@ -213,7 +213,11 @@ export class Extension implements RunHooks {
         if (!model)
           return vscode.window.showWarningMessage(messageNoPlaywrightTestsFound);
 
-        await this._reusedBrowser.record(model);
+        const project = model.enabledProjects()[0];
+        if (!project)
+          return vscode.window.showWarningMessage(this._vscode.l10n.t(`Project is disabled in the Playwright sidebar.`));
```
dgozman: Why can't we fallback to `model.config.testIdAttribute` like we do in `inspect()`?

mxschmitt: Do you mean something like https://github.com/microsoft/playwright-vscode/pull/644/commits/d46ca8b42f91b10dccd9636cce81fff975f8bfd1 instead of doing this?

```ts
return vscode.window.showWarningMessage(this._vscode.l10n.t(`Project is disabled in the Playwright sidebar.`));
```

src/extension.ts
```diff
@@ -198,22 +198,26 @@ export class Extension implements RunHooks {
...
           return;
 
-
+        const testIdAttributeName = project.project.use.testIdAttribute ?? project.model.config.testIdAttributeName;
         const showBrowser = this._settingsModel.showBrowser.get() ?? false;
         try {
           await this._settingsModel.showBrowser.set(true);
           await this._showBrowserForRecording(file, project);
-          await this._reusedBrowser.record(model);
+          await this._reusedBrowser.record(model, testIdAttributeName);
```
dgozman: I think we should pass `model, project` here, and just a `model` in recordAtCursor. Then, `ReusedBrowser` should have a helper method that will choose the right `testIdAttribute` and use that helper method in all the places. What I don't like in the current approach is the test id attribute logic scattered around the codebase. WDYT?

dgozman (approved):

PR #656, opened 2025-07-23
fix: don't run global setup for wrong project

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/36757.

`TestModel.narrowDownLocations` doesn't filter out test IDs from a completely different model. That means for a `TestRunRequest` against an individual test, we currently send it to every Test Server. This didn't surface because issuing a testrun with a test ID that doesn't exist on a given test server is sort of a no-op, but it also means we execute global setup on config that isn't part of the test run. Not a huge deal, but definitely wrong.

### DIFF
--- a/src/testModel.ts
+++ b/src/testModel.ts
@@ -651,15 +651,18 @@ export class TestModel extends DisposableBase {
       return { locations: [] };
     const locations = new Set<string>();
     const testIds: string[] = [];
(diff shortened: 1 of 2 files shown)


mxschmitt (approved):

danreeves:
Sorry for the ping but any chance we could get a release including this fix?

PR #657, opened 2025-07-24
chore: some UI sanding

Skn0tt:
- 12px instead of 4px on the left to align with styles of the tab above
- move "Update Snapshot" dropdown inline with label to save on vertical space

Before             |  After
:-------------------------:|:-------------------------:
  |

### DIFF
--- a/src/settingsView.ts
+++ b/src/settingsView.ts
@@ -258,37 +258,25 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
             <div>${vscode.l10n.t('Run global setup on each run')}</div>
           </label>
         </div class="action">
-        <div class="hbox">
-          <div class="action-indent"></div>
-          <div class="action-indent"></div>
-          <label id="updateSnapshotLabel">${vscode.l10n.t('Update snapshots')}</label>
-        </div>
-        <div class="combobox">
-          <div class="action-big-indent"></div>
-          <select setting="updateSnapshots" aria-labelledby="updateSnapshotLabel">
+        <label class="dropdown">
+          ${vscode.l10n.t('Update snapshots:')}
+          <select setting="updateSnapshots" class="combobox">
             <option value="no-override">no override</option>
             <option value="all">all</option>
             <option value="changed">changed</option>
             <option value="missing">missing</option>
             <option value="none">none</option>
           </select>
-          <div class="action-indent"></div>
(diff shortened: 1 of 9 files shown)


agg23:
It's difficult to tell what changed to the vertical spacing. Your indentation is technically an improvement, but because it's not aligned with the text of the header, it still doesn't look right.

Screenshot 2025-07-24 at 5 39 03 AM

Skn0tt:
as discussed offline, I aligned everything to the start and end of the vscode pane headers. 

Screenshot 2025-07-25 at 15 31 29

src/settingsView.ts
```diff
@@ -208,22 +208,22 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
       <title>Playwright</title>
     </head>
     <body class="settings-view">
-      <div class="section-header">${vscode.l10n.t('TOOLS')}</div>
+      <h2 class="section-header">${vscode.l10n.t('TOOLS')}</h2>
       <div id="actions" class="vbox"></div>
       <div id="model-selector" class="vbox" style="display: none">
-        <div class="section-header">
+        <h2 class="section-header">
```
agg23, line 214: Shouldn't this be a `h3`?

agg23 (commented):
Great improvement

src/settingsView.ts
Skn0tt: hmm, because the h1 is "VS Code", h2 is "Playwright" and this is h3? makes sense

Skn0tt: no, wait. why do you say it should be h3?

Skn0tt:
@dgozman this is the latest state: https://github.com/microsoft/playwright-vscode/pull/657#issuecomment-3117844137

src/settingsView.ts
agg23: I assume from the DOM that the second `h2` is embedded in the previous `h2`'s content. Thus it should step down a size (regardless of how it's styled, but it should likely be styled similarly).

Skn0tt: I see. It's not embedded in the previous's content - the previous is the section header for "actions", while this one is the section header for "projects". it's optional though, that's why it indents

Skn0tt:
I also updated the checkbox alignment.

Screenshot 2025-07-25 at 16 37 41

PR #664, opened 2025-08-01
test: unflake highlight test

mxschmitt:


tests/highlight-locators.spec.ts
```diff
@@ -35,7 +35,7 @@ test('should work', async ({ activate }) => {
     'test.spec.ts': `
       import { test } from '@playwright/test';
       test('one', async ({ page }) => {
-        await page.goto('https://example.com');
+        await page.goto('about:blank');
```
Skn0tt, line 38: if there's already the `setContent`, do we even need the goto?

Skn0tt (approved):

tests/highlight-locators.spec.ts
mxschmitt: We don't but I don't want to modify the rest of the test for now.

PR #666, opened 2025-08-01
feat(mcp): `browser_connect`

Skn0tt:
Adds logic for connecting `@playwright/mcp` to the reused browser. While it's detected and `Show Browser` is enabled, we ensure a reused browser server is running, and we send `browser_connect` to establish the connection.

This can be disabled in a couple of ways:
- via the "Connect Copilot" setting in the settings view, and in VS Code settings. It's disabled by default.
- via the "Show Browser" setting. If it's disabled, there's no reused browser server to connect to.
- when no MCP was detected. In that case, we also hide the "Connect Copilot" setting.

I've added tests that show how this plays together with our existing logic around starting and stopping a reused browser.

https://github.com/user-attachments/assets/a8ca2416-ac1d-4c5e-b45c-5414ec1f15a1

Depends on https://github.com/microsoft/playwright-mcp/pull/788.


src/reusedBrowser.ts
```diff
@@ -225,6 +235,24 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
     return this._recorderModeForTest;
   }
 
+  async connectMCP(mcp: McpConnection, model: TestModel) {
```
dgozman: I find this indirection undesirable. Let MCPConnection do all of this, and perhaps call `reusedBrowser.setKeepAlive(true/false)` to prevent auto-closure.

src/settingsView.ts
```diff
@@ -111,7 +115,15 @@ export class SettingsView extends DisposableBase implements vscodeTypes.WebviewV
   }
 
   private _updateSettings() {
-    void this._view!.webview.postMessage({ method: 'settings', params: { settings: this._settingsModel.json() } });
+    const settings = this._settingsModel.json();
+    const disabled = {
+      connectCopilot: !this._mcpConnection.isEnabled(),
+    };
+    const hidden = {
```
dgozman: Hidden UI is usually problematic. Let's instead disable and provide some information on hover, or upon an (i) button.

src/mcpConnection.ts
```diff
@@ -0,0 +1,84 @@
...
+    this._disposables.push(
+        new _vscode.Disposable(() => clearInterval(scanInterval)),
+        this._models.onUpdated(() => this._onUpdate.fire()),
+        this._reusedBrowser.onBackendChange(() => this._reconcile()),
+        this._settingsModel.connectCopilot.onChange(() => this._reconcile()),
+    );
+  }
+
+  private _reconcile() {
+    void this._commandQueue.run(async () => {
```
dgozman: You probably don't want to queue 10 operations when I click on the checkbox 10 times. Perhaps two flags `_needsUpdate` and `_isUpdating` would be better?

PR #668, opened 2025-08-14
feat: browser list UI

Skn0tt:
Adds UI for showing a list of browsers, each with a "Pick Locator" and "Close" button. This uses the data from https://github.com/microsoft/playwright/pull/37027. On old versions, we try to infer the browser information from the currently selected project.

While the buttons are per-browser in the UI, they currently always affect all browsers. I'll fix this in a follow-up, but it should be fine given most people only ever use a single browser at a time anways.

https://github.com/user-attachments/assets/4396c43f-580a-496c-8ad0-5230885c3303


package.nls.json
```diff
@@ -4,7 +4,8 @@
   "contributes.command.pw.extension.installBrowsers": "Install Playwright Browsers",
   "contributes.command.pw.extension.command.inspect": "Pick locator",
   "contributes.command.pw.extension.command.clearCache": "Clear cache",
-  "contributes.command.pw.extension.command.closeBrowsers": "Close all Playwright browsers",
+  "contributes.command.pw.extension.command.closeBrowsers": "Close all Playwright testing browsers",
```
dgozman: Are you going to follow up with translations? I wonder what is the process for that?

src/settingsView.ts
```diff
@@ -202,14 +264,20 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
...
     <head>
       <meta charset="UTF-8">
-      <meta http-equiv="Content-Security-Policy" content="default-src 'none'; style-src ${webview.cspSource}; script-src 'nonce-${nonce}';">
+      <meta http-equiv="Content-Security-Policy" content="default-src 'none'; style-src 'unsafe-inline' ${webview.cspSource}; script-src 'nonce-${nonce}';">
       <meta name="viewport" content="width=device-width, initial-scale=1.0">
       <link href="${style}" rel="stylesheet">
       <title>Playwright</title>
     </head>
     <body class="settings-view">
+      <section role="region" aria-label="Browsers">
```
dgozman, line 276: Perhaps the "Show browser" checkbox should be here as well?

src/settingsView.ts
```diff
@@ -134,8 +139,65 @@ export class SettingsView extends DisposableBase implements vscodeTypes.WebviewV
...
+            args: [b.id],
+          }
+        ]
+      };
+    });
+
+    if (browsers.length === 0) {
+      browsers.push({
+        svg: '',
+        text: 'No browsers open.',
```
dgozman: - `No open browsers`?
- l10n?

src/browserList.ts
```diff
@@ -0,0 +1,70 @@
...
+  _moderniseForTest = false;
+
+  private _onChanged: vscodeTypes.EventEmitter<void>;
+  readonly onChanged;
+
+  constructor(private readonly _vscode: vscodeTypes.VSCode, private readonly _reusedBrowser: ReusedBrowser, private readonly _models: TestModelCollection) {
+    this._onChanged = new this._vscode.EventEmitter();
+    this.onChanged = this._onChanged.event;
+
+    this._reusedBrowser.onBackend(b => this._add(b));
```
dgozman, line 52: Instead of listening to the `ReusedBrowser`, I'd make this a part of it. Or do you plan for this list to aggregate multiple backends?

src/common.ts
```diff
@@ -35,21 +35,32 @@ export interface ActionDescriptor {
   svg: string;
   title?: string;
   location?: string;
-  hidden?: boolean;
   disabled?: boolean;
 }
 
-export function createAction(action: ActionDescriptor, options?: { omitText?: boolean }): HTMLElement | null {
+export interface BrowserEntry {
```
dgozman, line 41: Shouldn't this be in `BrowserList`?

src/extension.ts
```diff
@@ -172,11 +175,14 @@ export class Extension implements RunHooks {
         for (const model of versions.values())
           await installBrowsers(this._vscode, model);
       }),
-      vscode.commands.registerCommand('pw.extension.command.inspect', async () => {
+      vscode.commands.registerCommand('pw.extension.command.inspect', async (browserId?: string) => {
```
dgozman, line 178: So this still works from the `Cmd+Shift+P` menu, enabling it for all browsers?

src/extension.ts
```diff
@@ -232,6 +238,26 @@ export class Extension implements RunHooks {
       vscode.commands.registerCommand('pw.extension.command.clearCache', async () => {
         await this._models.selectedModel()?.clearCache();
       }),
+      vscode.commands.registerCommand('pw.extension.command.closeBrowser', async (browserId?: string) => {
+        if (!browserId) {
+          const options = new Map<string, string>();
+          for (const b of this._browserList.get())
+            options.set(getBrowserTitle(b), b.id);
+          if (!options.size) {
+            await this._vscode.window.showErrorMessage('No browsers are open');
```
dgozman: Why show an error? If nothing's open, closing is an easy noop :)

src/extension.ts
```diff
@@ -232,6 +238,26 @@ export class Extension implements RunHooks {
...
+            return;
+          }
+          const selection = await this._vscode.window.showQuickPick([...options.keys()], { title: 'Select browser to close' });
```
dgozman: Hmm.... I'd close all browsers if running through `Cmd+Shift+P`, because the intention is probably to quickly close things. Perhaps name it "close all" as well?

src/settingsView.script.ts
```diff
@@ -47,16 +47,63 @@ function updateProjects(projects: ProjectEntry[]) {
   unselectAllButton.hidden = !allEnabled;
 }
 
+function updateBrowsers(browsers: BrowserEntry[]) {
+  const browsersElement = document.getElementById('browsers') as HTMLUListElement;
+  browsersElement.innerHTML = '';
```
dgozman: I always thought that setting `textContent` is the most efficient way.

```suggestion
  browsersElement.textContent = '';
```

src/browserList.ts
```diff
@@ -0,0 +1,70 @@
...
+      this._onChanged.fire();
+    });
+  }
+
+  get() {
+    return [...this._state.entries()].flatMap(([, state]) => state.browsers);
+  }
+}
+
+export function getBrowserTitle(browser: DebugControllerState['browsers'][0]): string {
```
dgozman: Why don't we calculate this in `stateChanged` event and have a convenient representation of the browser in the extension, instead of reusing whatever comes over `DebugController`?

src/settingsView.ts
```diff
@@ -134,8 +139,65 @@ export class SettingsView extends DisposableBase implements vscodeTypes.WebviewV
...
+      else if (b.name === 'webkit')
+        svg = webkitLogo;
+
+      return {
+        text: getBrowserTitle(b),
+        svg,
+        actions: [
+          {
+            svg: pickElementIcon,
+            title: 'Pick locator',
```
dgozman: l10n?

src/settingsView.ts
```diff
@@ -134,8 +139,65 @@ export class SettingsView extends DisposableBase implements vscodeTypes.WebviewV
...
+            command: 'pw.extension.command.inspect',
+            args: [b.id]
+          },
+          {
+            svg: closeIcon,
+            title: 'Close Browser',
```
dgozman: l10n?

src/settingsView.ts
```diff
@@ -134,8 +139,65 @@ export class SettingsView extends DisposableBase implements vscodeTypes.WebviewV
...
+    });
+
+    if (browsers.length === 0) {
+      browsers.push({
+        svg: '',
+        text: 'No browsers open.',
+        actions: [
+          {
+            svg: pickElementIcon,
+            title: 'Pick locator',
```
dgozman: l10n

src/settingsView.ts
```diff
@@ -134,8 +139,65 @@ export class SettingsView extends DisposableBase implements vscodeTypes.WebviewV
...
+            command: 'pw.extension.command.inspect',
+            args: []
+          },
+          {
+            svg: closeIcon,
+            title: 'Close Browser',
```
dgozman: l10n

src/settingsView.ts
```diff
@@ -202,14 +264,20 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
     <html lang="en">
     <head>
       <meta charset="UTF-8">
-      <meta http-equiv="Content-Security-Policy" content="default-src 'none'; style-src ${webview.cspSource}; script-src 'nonce-${nonce}';">
+      <meta http-equiv="Content-Security-Policy" content="default-src 'none'; style-src 'unsafe-inline' ${webview.cspSource}; script-src 'nonce-${nonce}';">
```
dgozman, line 270: Do you really need `unsafe-inline`?

package.nls.json
Skn0tt: Good question. I don't think we have a set process - let me follow up with translations at the end, when everything's said and done.

src/settingsView.ts
Skn0tt: perhaps! I'll try it out, but that's for a different PR.

src/browserList.ts
Skn0tt: Yep, the MCP server will also be contributing a DebugController, but with different behaviour than `ReusedBrowser`. BrowserList will aggregate both.

src/common.ts
Skn0tt: no, because it's part of the communication between `settingsView.ts` and `settingsView.script.ts`

src/extension.ts
Skn0tt: yes, <kbd>Cmd+Shift+P</kbd> sends `[]` args.

src/extension.ts
Skn0tt: because this command is for closing a single browser, whereas the other is for closing all of them. let me combine them though, you're right!

src/extension.ts
Skn0tt: yeah i'll combine them!

src/settingsView.ts
Skn0tt: yes, becaue the SVGs have `<def>` elements that don't work otherwise. I was unable to rewrite it so that it works without `unsafe-inline`. If you know a magic trick, I'd be happy to learn it!

src/browserList.ts
Skn0tt: good idea, done!

src/settingsView.ts
dgozman: Well, we can use a png icon, or find a different svg. [Here's an example](https://github.com/grommet/grommet-icons/blob/master/public/img/chrome.svg?short_path=692d202).

src/browserList.ts
```diff
@@ -0,0 +1,74 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { DebugController, DebugControllerState, ReusedBrowser } from './reusedBrowser';
+import { TestModelCollection } from './testModel';
+import * as vscodeTypes from './vscodeTypes';
+
+export class BrowserList {
+  private _state = new Map<DebugController, { id: string; name: string; channel?: string; title: string }[]>();
```
dgozman: I don't quite like that `BrowserList` touches `DebugController`, which is an internal part of it. I'd prefer for `BrowserList` to listen to `ReusedBrowser.onBrowsersChanged` event instead, and encapsulate the `DebugController` inside. WDYT?

Skn0tt: I think it's good how it is. Let me give an outlook for what i'm planning in the upcoming PRs:

`ReusedBrowser` won't be the only thing that contributes Debug Controllers, but the MCP integration will also contribute them. So I want all browser-related actions to live on `BrowserList` instead of `ReusedBrowser`. For that, I need DebugController to become an internal part of `BrowserList`. `ReusedBrowser` will lose a lot of responsibility to `BrowserList`, all the inspection and closing and recording stuff will move over to `BrowserList`, and `ReusedBrowser` ends up being just about starting/stopping the browser server.

Skn0tt: As discussed offline, I updated this PR so that it moves over the DebugController-related parts right away.

PR #670, opened 2025-08-21
chore: sort paths intuitively

Skn0tt:
I noticed that in the `playwright-mcp` repo, the extension selected `extension/playwright.config.ts` as the default config, instead of the more intuitive `playwright.config.ts`. By taking path depth into account when sorting, not just lexicals, we get a more intuitive ordering, and will select `playwright.config.ts` as the default.

### DIFF
--- a/tests/track-configs.spec.ts
+++ b/tests/track-configs.spec.ts
@@ -128,3 +128,16 @@ test('should show config loading errors', async ({ vscode, activate }) => {
   void testController.run(testItems);
   await expect.poll(() => vscode.window.activeTextEditor?.document.uri.toString()).toContain('playwright1.config.js');
 });
+
+test('should select root config by default', async ({ activate }) => {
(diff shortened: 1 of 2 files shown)


mxschmitt (approved):
Ich habe nichts dagegen einzuwenden.

tests/track-configs.spec.ts
```diff
@@ -128,3 +129,32 @@ test('should show config loading errors', async ({ vscode, activate }) => {
...
+        return items;
+      };
+    });
+
+    await webView.getByTitle('Toggle Playwright Configs').click();
+
+    return items;
+  }).toEqual([
+    'playwright.config.ts',
+    'extension/playwright.config.ts'
```
mxschmitt:
```suggestion
    `extension${path.sep}playwright.config.ts`
```

PR #671, opened 2025-08-22
chore: align test-server `PATH` with `npx playwright test`

Skn0tt:
We recommend running Playwright via `npx`, which brings all project dependencies into `PATH`. This means that in a webserver config, `serve build` will run the binary from `node_modules/.bin/serve`. Some projects in the wild make use of this (https://github.com/DerYeger/yeger, https://github.com/louislam/uptime-kuma), and so our extension fails spinning up a webserver that works just fine with `npx playwright test`.

To fix this, this PR adds env that resembles npx, loosely based on https://github.com/sindresorhus/npm-run-path/blob/main/index.js.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -140,25 +140,27 @@ async function findNodeViaShell(vscode: vscodeTypes.VSCode, cwd: string): Promis
   });
 }
 
-export async function findNpx(vscode: vscodeTypes.VSCode, cwd: string): Promise<string | undefined> {
-  const node = await findNode(vscode, cwd);
-  const npxPath = path.resolve(node, '..', 'npx');
-  try {
-    const stat = await fs.promises.stat(npxPath);
-    if (!stat.isFile())
-      return;
+export function addNpmRunPath(env: NodeJS.ProcessEnv, cwd: string): NodeJS.ProcessEnv {
+  const newPath = [];
 
-    if (process.platform !== 'win32') {
-      try {
-        await fs.promises.access(npxPath, fs.constants.X_OK);
-      } catch {
-        return;
-      }
-    }
-
-    return npxPath;
-  } catch {
+  let currentPath = path.resolve(cwd);
+  let previousPath;
+  while (previousPath !== currentPath) {
+    newPath.push(path.join(currentPath, 'node_modules', '.bin'));
+    previousPath = currentPath;
+    currentPath = path.resolve(currentPath, '..');
   }
+
+  // On Windows, PATH key casing can be “Path”; preserve whichever exists
+  const defaultPath = process.platform === 'win32' ? 'Path' : 'PATH';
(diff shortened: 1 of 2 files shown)


pavelfeldman:
This would be a substantial slowdown, especially on Windows. Is there a bug filed for this?

src/extension.ts
```diff
@@ -368,7 +368,7 @@ export class Extension implements RunHooks {
       if (error) {
         if (error.location) {
           const document = await this._vscode.workspace.openTextDocument(error.location.file);
-          const position = new this._vscode.Position(error.location.line - 1, error.location.column - 1);
+          const position = new this._vscode.Position(Math.max(0, error.location.line - 1), error.location.column - 1);
```
dgozman, line 371: These changes seem unrelated.

src/utils.ts
```diff
@@ -140,6 +140,29 @@ async function findNodeViaShell(vscode: vscodeTypes.VSCode, cwd: string): Promis
   });
 }
 
+export function addNpmRunPath(env: NodeJS.ProcessEnv, cwd: string): NodeJS.ProcessEnv {
+  const newPath = [];
+
+  let currentPath = path.resolve(cwd);
+  let previousPath;
+  while (previousPath !== currentPath) {
+    newPath.push(path.join(currentPath, 'node_modules', '.bin'));
```
dgozman, line 149: Shouldn't we check that `node_modules/.bin` exists? Looks kinda wasteful to add all of them.

dgozman (approved):

src/extension.ts
Skn0tt: They're unrelated, but I stumbled into weird errors here while working on this change. The `Position` constructor threw because of `-1` being passed. Let me try reproing them.

src/utils.ts
Skn0tt: I'm not sure! My hunch is that the system that resolves `PATH` is *very* optimised and much faster at checking path existence than our code. And because FS operations come at a premium under Windows, leaving the actual checks to the OS would be faster. This is all speculation though - do you happen to know more about this?

dgozman: No, I don't. I see that's what the other project does. Let's keep it as is.

src/extension.ts
Skn0tt: I think it's from here: https://github.com/Skn0tt/playwright-vscode/blob/cfb259a1015be3cd112ebcef3339e0d689f0652a/src/testModel.ts#L188

Skn0tt: There's several places where Playwright emits `line: 0, column: 0`: https://github.com/search?q=repo%3Amicrosoft%2Fplaywright%20line%3A%200&type=code
I'll kepp this in.

PR #672, opened 2025-08-22
chore(ui): hide projects section when there's a single project

Skn0tt:
Some Playwright workspaces don't define projects, or don't define project names. We show a single project named `<untitled>`. That UI is confusing and wastes space. This PR hides the projects section when there's only a single project.

### DIFF
--- a/src/settingsView.script.ts
+++ b/src/settingsView.script.ts
@@ -23,6 +23,13 @@ const unselectAllButton = document.getElementById('unselectAll') as HTMLAnchorEl
 const toggleModels = document.getElementById('toggleModels') as HTMLAnchorElement;
 
 function updateProjects(projects: ProjectEntry[]) {
+  const projectSelector = document.getElementById('project-selector')!;
+  if (projects.length < 2) {
+    projectSelector.style.display = 'none';
+    return;
+  }
+  projectSelector.style.display = 'flex';
+
   const projectsElement = document.getElementById('projects') as HTMLElement;
   projectsElement.textContent = '';
   for (const project of projects) {
--- a/src/settingsView.ts
+++ b/src/settingsView.ts
@@ -223,18 +223,20 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
           <select data-testid="models" id="models" title="${vscode.l10n.t('Select Playwright Config')}" ></select>
         </div>
       </div>
-      <h2 class="section-header">
-        ${vscode.l10n.t('PROJECTS')}
-        <div class="section-toolbar">
-          <a id="selectAll" role="button" title="${vscode.l10n.t('Select All')}">
-            <svg width="48" height="48" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" fill="currentColor"><path d="M9 9H4v1h5V9z"/><path d="M7 12V7H6v5h1z"/><path fill-rule="evenodd" clip-rule="evenodd" d="M5 3l1-1h7l1 1v7l-1 1h-2v2l-1 1H3l-1-1V6l1-1h2V3zm1 2h4l1 1v4h2V3H6v2zm4 1H3v7h7V6z"/></svg>
-          </a>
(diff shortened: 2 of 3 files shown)


src/settingsView.script.ts
```diff
@@ -23,6 +23,13 @@ const unselectAllButton = document.getElementById('unselectAll') as HTMLAnchorEl
 const toggleModels = document.getElementById('toggleModels') as HTMLAnchorElement;
 
 function updateProjects(projects: ProjectEntry[]) {
+  const projectSelector = document.getElementById('project-selector')!;
+  if (projects.length < 2) {
+    projectSelector.style.display = 'none';
```
dgozman, line 28: What if I have a single unchecked project? How will I be able to check it?

Skn0tt: Without prior state, that single project will be checked by default and because the UI is hidden you won't be able to uncheck it. You're worried about existing users with prior state, right?

dgozman: I guess so. I was also not aware that we check the single project by default, that's good to know.

Skn0tt: here's where we do it: https://github.com/Skn0tt/playwright-vscode/blob/c71ce8b1636383eb1d5899e3200e1a32747104b0/src/testModel.ts#L103-L116

I'll try adding a clause the ensures the single project is checked, for users with prior state.

Skn0tt: Done in https://github.com/microsoft/playwright-vscode/pull/672/commits/12df99dd66d98ecb26d80285572d0ec4fbd31c94. When the projects UI is hidden, we enable all projects of the selected model.

src/settingsView.ts
```diff
@@ -157,7 +157,19 @@ export class SettingsView extends DisposableBase implements vscodeTypes.WebviewV
       });
     }
 
-    void this._view.webview.postMessage({ method: 'models', params: { configs, showModelSelector: this._models.models().length > 1 } });
+    const showProjectSelector = this._models.enabledModels().some(model => model.projects().length > 1);
+    const selectedModel = this._models.selectedModel();
+    if (!showProjectSelector && selectedModel?.enabledProjects().length === 0)
+      this._models.setAllProjectsEnabled(selectedModel.config.configFile, true);
```
dgozman: This looks like an arbitrary place to select projects. I think we can instead update `enabledProjects()` to always consider the single project to be enabled. WDYT?

Skn0tt: sounds good!

Skn0tt:
I discussed the UI jumping with Pavel offline, and we agreed that it's fine for the UI under the control to jump. I'll revert the changes about changing when the UI is shown.

src/testModel.ts
```diff
@@ -156,12 +156,18 @@ export class TestModel extends DisposableBase {
     return [...new Set([...this._projects.values()].map(p => p.project.testDir))];
   }
 
+  isProjectEnabled(project: TestProject) {
+    if (this._projects.size < 2)
+      return true;
+    return project.isEnabled;
```
dgozman: Can we make `project.isEnabled` private or hide it in some other ways, so that no other place mistakenly uses it?

Skn0tt: The only way I could think if is with a symbol: https://github.com/microsoft/playwright-vscode/pull/672/commits/4de8c0c68131ad1f407bbd9b32f055f516ccae3a

src/testModel.ts
```diff
@@ -850,7 +858,7 @@ export class TestModelCollection extends DisposableBase {
         relativeConfigFile: path.relative(model.config.workspaceFolder, model.config.configFile),
         selected: model.config.configFile === this._selectedConfigFile,
         enabled: model.isEnabled,
-        projects: model.projects().map(p => ({ name: p.name, enabled: p.isEnabled })),
+        projects: model.projects().map(p => ({ name: p.name, enabled: model.isProjectEnabled(p)})),
```
dgozman: I think this should be `p[kIsEnabled]`. We want to save the explicit user values instead of our own calculation. This way it is preserved when adding/removing projects in the config.

Skn0tt: Good catch, I need coffee.

PR #673, opened 2025-08-22
test: add integration test for pnpm

Skn0tt:


mxschmitt (approved):
Having it with the `node-linker=pnp` mode would be interesting!

PR #675, opened 2025-08-25
chore: yarn PnP support

Skn0tt:
I'm going through open source Playwright projects, and found a couple that use Yarn PnP or PNPM. I'd like to make sure that they work with the extension by default, but PnP currently doesn't. This PR implements the workaround from https://github.com/microsoft/playwright/issues/18931. PnP is complicated, so I'm expecting us to invite a ton of bugs with this. Let's take that complexity from our users and load it onto ourselves.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -350,9 +350,9 @@ export class Extension implements RunHooks {
       return typeof entry[1] === 'string' ? entry : [entry[0], JSON.stringify(entry[1])];
     })) as NodeJS.ProcessEnv;
 
-    if (configFile && this._pnpFiles.has(configFile)) {
+    if (!env.NODE_OPTIONS && this._pnpFiles.has(configFile)) {
       const { pnpCJS, pnpLoader } = this._pnpFiles.get(configFile)!;
-      env.NODE_OPTIONS ??= '';
+      env.NODE_OPTIONS = '';
       if (pnpCJS)
         env.NODE_OPTIONS += ` --require ${pnpCJS}`;
       if (pnpLoader)
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -350,14 +350,18 @@ export class Extension implements RunHooks {
       return typeof entry[1] === 'string' ? entry : [entry[0], JSON.stringify(entry[1])];
     })) as NodeJS.ProcessEnv;
 
-    if (!env.NODE_OPTIONS && this._pnpFiles.has(configFile)) {
-      const { pnpCJS, pnpLoader } = this._pnpFiles.get(configFile)!;
-      env.NODE_OPTIONS = '';
-      if (pnpCJS)
-        env.NODE_OPTIONS += ` --require ${pnpCJS}`;
-      if (pnpLoader)
-        env.NODE_OPTIONS += ` --experimental-loader ${pathToFileURL(pnpLoader)}`;
-    }
+    if (env.NODE_OPTIONS?.includes('.pnp.cjs') || env.NODE_OPTIONS?.includes('.pnp.js'))
+      return env;
+
+    if (!this._pnpFiles.has(configFile))
+      return env;
+
+    env.NODE_OPTIONS ??= '';
+    const { pnpCJS, pnpLoader } = this._pnpFiles.get(configFile)!;
+    if (pnpCJS)
+      env.NODE_OPTIONS += ` --require ${pnpCJS}`;
+    if (pnpLoader)
+      env.NODE_OPTIONS += ` --experimental-loader ${pathToFileURL(pnpLoader)}`;
 
     return env;
   }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -369,8 +369,8 @@ export class Extension implements RunHooks {
   private async _detectPnp(configFileUri: vscodeTypes.Uri, root: string) {
     let dir = this._vscode.Uri.joinPath(configFileUri, '..');
     while (uriToPath(dir).length >= root.length) {
-      const pnpCjs = this._vscode.Uri.joinPath(configFileUri, '..', '.pnp.cjs');
-      const pnpLoader = this._vscode.Uri.joinPath(configFileUri, '..', '.pnp.loader.mjs');
+      const pnpCjs = this._vscode.Uri.joinPath(dir, '.pnp.cjs');
+      const pnpLoader = this._vscode.Uri.joinPath(dir, '.pnp.loader.mjs');
       const [pnpCjsExists, pnpLoaderExists] = await Promise.all([
         this._fileExists(pnpCjs),
         this._fileExists(pnpLoader)
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -367,8 +367,9 @@ export class Extension implements RunHooks {
   }
 
   private async _detectPnp(configFileUri: vscodeTypes.Uri, root: string) {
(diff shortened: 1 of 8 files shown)


src/extension.ts
```diff
@@ -339,11 +339,32 @@ export class Extension implements RunHooks {
...
+    return this._enrichWithPNP(env);
+  }
+
+  private async _enrichWithPNP(env: NodeJS.ProcessEnv): Promise<NodeJS.ProcessEnv> {
+    if (env.NODE_OPTIONS)
+      return env;
+
+    // based on the workaround from https://github.com/microsoft/playwright/issues/18931
+    // TODO: support multiple .pnp.cjs files; always pick the one closest to the config file.
+    const pnpFile = await this._vscode.workspace.findFiles('**/.pnp.cjs', null, 1);
```
pavelfeldman: Nit: You probably want to push it up into async section instead of making env async.

Skn0tt:
This doesn't yet work with PNPM's PnP support due to a bug on their end. It's fixed now and should be released soon, once that's done I can add tests for PNPM's PnP as well. https://github.com/pnpm/pnpm/issues/9904

src/extension.ts
```diff
@@ -339,11 +344,55 @@ export class Extension implements RunHooks {
...
-  private _envProvider(): NodeJS.ProcessEnv {
-    const env = this._vscode.workspace.getConfiguration('playwright').get('env', {});
-    return Object.fromEntries(Object.entries(env).map(entry => {
+  private _envProvider(configFile: string) {
+    const config = this._vscode.workspace.getConfiguration('playwright').get('env', {});
+    const env = Object.fromEntries(Object.entries(config).map(entry => {
       return typeof entry[1] === 'string' ? entry : [entry[0], JSON.stringify(entry[1])];
     })) as NodeJS.ProcessEnv;
+
+    if (!env.NODE_OPTIONS && this._pnpFiles.has(configFile)) {
```
dgozman: Why checking for `!env.NODE_OPTIONS`? Perhaps there's something unrelated there?

src/extension.ts
```diff
@@ -339,11 +344,55 @@ export class Extension implements RunHooks {
...
+      }
+
+      dir = this._vscode.Uri.joinPath(dir, '..');
+    }
+  }
+
+  private async _fileExists(uri: vscodeTypes.Uri) {
+    try {
+      await this._vscode.workspace.fs.stat(uri);
+      return true;
```
dgozman:
```suggestion
      const stat = await this._vscode.workspace.fs.stat(uri);
      return !!(stat.type & this._vscode.FileType.File);
```

src/extension.ts
```diff
@@ -339,11 +344,55 @@ export class Extension implements RunHooks {
...
+      if (pnpLoader)
+        env.NODE_OPTIONS += ` --experimental-loader ${pathToFileURL(pnpLoader)}`;
+    }
+
+    return env;
+  }
+
+  private async _detectPnp(configFileUri: vscodeTypes.Uri, root: string) {
+    let dir = this._vscode.Uri.joinPath(configFileUri, '..');
+    while (uriToPath(dir).length >= root.length) {
```
dgozman: We usually check `while (dir != root)` instead of length. Or `while (path.resolve(dir, '..') !== dir)`. Does that fail here?

src/extension.ts
```diff
@@ -339,11 +344,55 @@ export class Extension implements RunHooks {
...
+      const pnpCjs = this._vscode.Uri.joinPath(configFileUri, '..', '.pnp.cjs');
```
dgozman: Shouldn't this use `dir` instead of `configFileUri`? Same in the next line.

tests/pnp.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+test('should pick up .pnp.cjs file closest to config', async ({ activate }, testInfo) => {
+  const pnpCjsOutfile = testInfo.outputPath('pnp-cjs.txt');
+  const esmLoaderInfile = testInfo.outputPath('esm-loader-in.txt');
+  const esmLoaderOutfile = testInfo.outputPath('esm-loader-out.txt');
+  await fs.writeFile(esmLoaderInfile, 'untouched');
+
+  const { vscode, testController } = await activate({
+    'playwright.config.js': `module.exports = { testDir: 'tests' }`,
+    '.pnp.cjs': `
```
dgozman, line 28: Let's put this file one/two directories up from the config to exercise the path traversal logic.

src/extension.ts
Skn0tt: The idea is that some users have this exact workaround in their config, and I don't want NODE_OPTIONS to end up as `--require=.pnp.cjs --require=.pnp.cjs`. Not sure what'd happen then.

But maybe `NODE_OPTIONS.includes('.pnp.cjs') || NODE_OPTIONS.includes('.pnp.js')` is good enough of a heuristic? It feels brittle but should probably work.

dgozman: I like the heurisitc!

src/extension.ts
Skn0tt: `dir != root` only works if `uriToPath()` normalises trailing slashes in the same way that `root` is. I'm not sure about that because there's VS Code APIs in the mix, and not just node.js stdlib. I'll look into that. Not sure about `path.resolve(dir, '..') !== dir` - wouldn't that only stop at `/`, instead of the workspace root?

src/extension.ts
```diff
@@ -353,23 +353,28 @@ export class Extension implements RunHooks {
...
 
-    if (!env.NODE_OPTIONS && this._pnpFiles.has(configFile)) {
-      const { pnpCJS, pnpLoader } = this._pnpFiles.get(configFile)!;
-      env.NODE_OPTIONS = '';
-      if (pnpCJS)
-        env.NODE_OPTIONS += ` --require ${pnpCJS}`;
-      if (pnpLoader)
-        env.NODE_OPTIONS += ` --experimental-loader ${pathToFileURL(pnpLoader)}`;
-    }
+    if (env.NODE_OPTIONS?.includes('.pnp.cjs') || env.NODE_OPTIONS?.includes('.pnp.js'))
```
dgozman: Why do we have different filenames here and in `_detectPnp`?

src/extension.ts
```diff
@@ -384,15 +389,13 @@ export class Extension implements RunHooks {
...
-
-      dir = this._vscode.Uri.joinPath(dir, '..');
     }
   }
 
   private async _fileExists(uri: vscodeTypes.Uri) {
     try {
-      await this._vscode.workspace.fs.stat(uri);
-      return true;
+      const stat = await this._vscode.workspace.fs.stat(uri);
```
dgozman: I looked around, and we never use `workspace.fs`. Instead, we use regular `fs` all the time. Perhaps we can do with regular `path`/`fs` instead of vscode's `uri` and `workspace`?

Skn0tt: I looked it up: https://code.visualstudio.com/api/extension-guides/virtual-workspaces

We'll never be able to support virtual workspaces since we need child processes, so you're right we can just use regular path/fs.

src/extension.ts
Skn0tt: Because I was somehow under the impression that the `.pnp.loader.js` file never comes without `.pnp.cjs`, but that's not a given. I've updated the condition to match the other filenames.

PR #676, opened 2025-08-25
chore: on headless linux, debug headlessly

Skn0tt:
Resolves https://github.com/microsoft/playwright/issues/23805.


Skn0tt:
Currently, when debugging a test on Codespaces, we just throw an error. Not throwing an error is arguably better - but it's still not the most useful thing in the world, agreed.

PR #679, opened 2025-08-27
test: unskip yarn-berry windows integration tests

mxschmitt:
Otherwise it ends up in a

```
'C:\.tools\.npm-global\bin\create-playwright@latest' is not recognized as an internal or external command,
operable program or batch file.
error Command failed.
Exit code: 1
```


Skn0tt (approved):
nice!

PR #680, opened 2025-08-27
chore: drop "Running global setup" line

Skn0tt:
This line was added as an unrelated change in https://github.com/microsoft/playwright-vscode/pull/478. It's printed on the first test run, but unconditionally. Users without plugins or global setup will still see it, and might be confused. I think it's better to leave it out - if the global setup file logs something, it'll be visible, otherwise it's fine to not see it.

### DIFF
--- a/src/upstream/teleEmitter.ts
+++ b/src/upstream/teleEmitter.ts
@@ -162,6 +162,8 @@ export class TeleReporterEmitter implements ReporterV2 {
       rootDir: config.rootDir,
       version: config.version,
       workers: config.workers,
+      globalSetup: config.globalSetup,
+      globalTeardown: config.globalTeardown,
     };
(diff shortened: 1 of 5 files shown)


dgozman:
I remember that users expressed confusion about whether the global setup runs or not. Let's think about the ways to make this easier to understand?

src/upstream/teleEmitter.ts
```diff
@@ -162,6 +162,8 @@ export class TeleReporterEmitter implements ReporterV2 {
       rootDir: config.rootDir,
       version: config.version,
       workers: config.workers,
+      globalSetup: config.globalSetup,
```
dgozman, line 165: Let's follow up with mirroring this upstream.

PR #684, opened 2025-09-01
chore: show mcp-owned browsers

Skn0tt:
Scans `vscode.lm.tools` for Playwright MCP instances, and calls the `browser_connect` tool to establish a DebugController connection with them. https://github.com/microsoft/playwright-mcp/pull/979 shows its implementation.
These debug controllers can then be shown from the browser list, just like the test-runner controlled debug controllers.

### DIFF
--- a/src/reusedBrowser.ts
+++ b/src/reusedBrowser.ts
@@ -116,9 +116,8 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
         this._testingBackend = undefined;
       }
     });
-    backend.onError(e => {
+    backend.onError(() => {
       if (backend === this._testingBackend) {
-        void this._vscode.window.showErrorMessage(e.message);
         void this._reset('none', this._testingBackend);
         this._testingBackend = undefined;
       }


src/reusedBrowser.ts
```diff
@@ -125,14 +125,38 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
...
+    this._addDebugController(backend, model);
+    return {};
+  }
+
+  async connectToDebugController(wsEndpoint: string, version: number, onClose: () => void) {
+    const backend = new Backend(this._vscode, version);
+    backend.onClose(onClose);
+    backend.onError(onClose);
+    this._addDebugController(backend);
+    // connect after attaching listeners, so we get the initial state event
```
dgozman, line 136: I wonder why we do it differently for the regular `backend`.

src/reusedBrowser.ts
```diff
@@ -26,14 +26,14 @@ type RecorderMode = 'none' | 'standby' | 'inspecting' | 'recording';
...
+  private _backend: Backend | undefined; // test runner backend
   private _cancelRecording: (() => void) | undefined;
   private _isRunningTests = false;
   private _insertedEditActionCount = 0;
   private _envProvider: () => NodeJS.ProcessEnv;
   private _disposables: vscodeTypes.Disposable[] = [];
-  private _pageCount = 0;
-  private _openBrowsers: { id?: string; name: string; channel?: string; title: string }[] = [];
+  private _pageCount = 0; // test runner pagecount
+  private _debugControllers = new Map<Backend, { id?: string; name: string; channel?: string; title: string }[]>();
```
dgozman: I think `_openBrowsers` was a better name for this one.

src/reusedBrowser.ts
```diff
@@ -26,14 +26,14 @@ type RecorderMode = 'none' | 'standby' | 'inspecting' | 'recording';
...
-  private _backend: Backend | undefined;
+  private _backend: Backend | undefined; // test runner backend
   private _cancelRecording: (() => void) | undefined;
   private _isRunningTests = false;
   private _insertedEditActionCount = 0;
   private _envProvider: () => NodeJS.ProcessEnv;
   private _disposables: vscodeTypes.Disposable[] = [];
-  private _pageCount = 0;
-  private _openBrowsers: { id?: string; name: string; channel?: string; title: string }[] = [];
+  private _pageCount = 0; // test runner pagecount
```
dgozman: `_testingPageCount`?

src/reusedBrowser.ts
```diff
@@ -26,14 +26,14 @@ type RecorderMode = 'none' | 'standby' | 'inspecting' | 'recording';
 
 export class ReusedBrowser implements vscodeTypes.Disposable {
   private _vscode: vscodeTypes.VSCode;
-  private _backend: Backend | undefined;
+  private _backend: Backend | undefined; // test runner backend
```
dgozman: `_testingBackend`?

src/reusedBrowser.ts
```diff
@@ -178,12 +202,18 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
...
             title,
           };
-        });
+        }));
       }
 
-      this._pageCountChanged(params.pageCount);
+      this._onPageCountChangedEvent.fire(this.pageCount());
+
+      if (backend === this._backend) {
```
dgozman: Why do we need this logic now? It was not present before.

src/reusedBrowser.ts
```diff
@@ -266,32 +313,27 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
...
+      try {
+        await backend.setRecorderMode({
+          mode: 'inspecting',
+          testIdAttributeName,
+          browserId,
+        });
+        this._recorderModeForTest = 'inspecting';
+      } catch (e) {
+        showExceptionAsUserError(this._vscode, selectedModel, e as Error);
+        return;
```
dgozman: Perhaps continue with other backends instead of returning?

src/mcpConnection.ts
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+
+import { ReusedBrowser } from './reusedBrowser';
+import { VSCode } from './vscodeTypes';
+
+export class McpConnection {
+  private _knownEndpoints = new Set<string>();
+
+  constructor(private _vscode: VSCode, private _reusedBrowser: ReusedBrowser) {}
```
dgozman: As discussed before, either all fields are declared here, or all fields are declared explicitly above, but not mixing both.

src/mcpConnection.ts
```diff
@@ -0,0 +1,53 @@
...
+    const timeout = setTimeout(() => this._checkForUpdates().finally(() => timeout.refresh()), 500);
+    return new this._vscode.Disposable(() => clearTimeout(timeout));
+  }
+
+  private async _checkForUpdates(): Promise<void> {
+    for (const tool of this._vscode.lm.tools) {
+      if (!tool.name.endsWith('browser_connect'))
+        continue;
+
+      const result = await this._vscode.lm.invokeTool(tool.name, {
```
dgozman, line 43: Can we avoid calling the same tool every 500ms over and over again?

src/mcpConnection.ts
```diff
@@ -0,0 +1,53 @@
...
+        toolInvocationToken: undefined,
+        input: { debugController: true }
+      });
+      const text = result.content.map(c => '' + c).join('\n');
+      const url = text.match(/URL: (.+)/)?.[1];
+      const version = parseFloat(text.match(/Version: (\d+\.\d+)/)?.[1] ?? 'NaN');
```
dgozman, line 55: An example of the tool output in the comment would be handy here.

src/mcpConnection.ts
```diff
@@ -0,0 +1,53 @@
...
+      if (!url || isNaN(version) || this._knownEndpoints.has(url))
+        continue;
+
+      this._knownEndpoints.add(url);
+      await this._reusedBrowser.connectToDebugController(
```
dgozman, line 67: How do we disconnect from it? A comment explaining the lifetime will help.

src/reusedBrowser.ts
Skn0tt: I was hoping you'd say that :D Didn't do it originally to keep the diff crisp.

src/reusedBrowser.ts
Skn0tt: We had it, it was in `this._pageCountChanged()`.

src/reusedBrowser.ts
Skn0tt: The regular backend starts off empty, so the initial state event isn't interesting. This is different for MCP.

src/mcpConnection.ts
Skn0tt: 😢

src/reusedBrowser.ts
```diff
@@ -111,28 +111,52 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
     if (!backend)
       return { errors };
     backend.onClose(() => {
-      if (backend === this._backend) {
-        this._backend = undefined;
+      if (backend === this._testingBackend) {
+        this._testingBackend = undefined;
         void this._reset('none');
```
dgozman: I think this logic should later apply to the "currently recording backend", once you implement recording in all backends.

src/reusedBrowser.ts
```diff
@@ -111,28 +111,52 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
...
       }
     });
     backend.onError(e => {
-      if (backend === this._backend) {
+      if (backend === this._testingBackend) {
         void this._vscode.window.showErrorMessage(e.message);
-        this._backend = undefined;
+        this._testingBackend = undefined;
         void this._reset('none');
```
dgozman: Same here.

src/reusedBrowser.ts
```diff
@@ -82,7 +82,7 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
 
   private async _startBackendIfNeeded(model: TestModel): Promise<{ errors?: string[] }> {
     // Unconditionally close selector dialog, it might send inspect(enabled: false).
-    if (this._backend) {
+    if (this._testingBackend) {
```
dgozman, line 85: Let's follow up with making this piece understandable, as discussed offline.

src/reusedBrowser.ts
```diff
@@ -441,14 +486,14 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
   private async _reset(mode: 'none' | 'standby') {
     this._resetExtensionState();
     this._recorderModeForTest = mode;
-    await this._backend?.setRecorderMode({ mode });
+    await this._testingBackend?.setRecorderMode({ mode });
```
dgozman: Only touching the testing backend here seems suspicious.

src/reusedBrowser.ts
```diff
@@ -178,12 +202,18 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
...
-        });
+        }));
       }
 
-      this._pageCountChanged(params.pageCount);
+      this._onPageCountChangedEvent.fire(this.pageCount());
+
+      if (backend === this._testingBackend) {
+        this._testingPageCount = params.pageCount;
+        if (!this._isRunningTests && this._testingPageCount === 0)
```
dgozman: Let's extract a method `_maybeStopTestingBackend()` and call it here, and from `didRunTests()`.

src/reusedBrowser.ts
```diff
@@ -111,33 +111,57 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
...
+        this._testingBackend = undefined;
       }
     });
     backend.onError(e => {
-      if (backend === this._backend) {
+      if (backend === this._testingBackend) {
         void this._vscode.window.showErrorMessage(e.message);
-        this._backend = undefined;
-        void this._reset('none');
+        void this._reset('none', this._testingBackend);
```
dgozman, line 121: It feels like this call should be inside the handler on the line 147 instead. Same goes for lines 115 and 143.

src/reusedBrowser.ts
```diff
@@ -232,25 +285,18 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
   }
 
   pageCount() {
-    return this._pageCount;
+    let sum = 0;
+    for (const browsers of this._openBrowsers.values())
+      sum += browsers.length;
```
dgozman: Rename to `browserCount()` or sum pages instead?

src/reusedBrowser.ts
Skn0tt: For `onError`, maybe. But not for `onClose`, since the extension doesn't control the MCP lifecycle. I'll revisit this once I implement recording for MCP.

PR #685, opened 2025-09-02
chore(reused browser): close after test run finishes

Skn0tt:
Unchecking "Show Browser" during a test run changes the setting in VS Code's store, but then shows a warning and doesn't actually stop the browser server:

https://github.com/microsoft/playwright-vscode/blob/6663081da3e1c925d6a46764796acb422092abcc/src/reusedBrowser.ts#L426-L431

On the next run, the browser server continues to be used, even. This is surprising behaviour. We shouldn't take away the browser server from an ongoing test run, but at least we can stop it afterwards. That's what his PR does, plus some cleanup.

### DIFF
--- a/src/playwrightTestServer.ts
+++ b/src/playwrightTestServer.ts
@@ -47,13 +47,7 @@ export type PlaywrightTestRunOptions = {
   updateSourceMethod?: 'overwrite' | 'patch' | '3way' | undefined;
 };
 
-export interface RunHooks {
-  onWillRunTests(model: TestModel, debug: boolean): Promise<{ connectWsEndpoint?: string }>;
-  onDidRunTests(debug: boolean): Promise<void>;
-}
-
 export type PlaywrightTestOptions = {
-  runHooks: RunHooks;
   isUnderTest: boolean;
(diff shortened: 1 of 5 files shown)


src/reusedBrowser.ts
```diff
@@ -400,8 +400,8 @@ export class ReusedBrowser implements vscodeTypes.Disposable {
     await this._startBackendIfNeeded(model);
   }
 
-  async onDidRunTests(debug: boolean) {
-    if (debug && !this._settingsModel.showBrowser.get()) {
```
Skn0tt, line 404: this is the core change

src/playwrightTestServer.ts
```diff
@@ -362,7 +356,6 @@ export class PlaywrightTestServer {
       if (!token.isCancellationRequested && debugTestServer && !debugTestServer.isClosed())
         await this._runGlobalHooksInServer(debugTestServer, 'teardown', reporter, token);
       debugTestServer?.close();
-      await this._options.runHooks.onDidRunTests(true);
```
Skn0tt, line 365: we're already making this call in `testModel.ts`, no need to do it here

PR #690, opened 2025-09-09
chore: dont enable just setup project by default

Skn0tt:
While going through VS Code extension feedback, I came across a review linking to https://github.com/microsoft/playwright/issues/35731. This PR resolves that.

When opening a Playwright project for the first time, we enable the *first* project in the project list. In the case of this issue, that's a `setup` project that's not useful alone. Putting the setup project first is [what we recommend in our docs](https://playwright.dev/docs/test-global-setup-teardown#setup), so I reckon this isn't the only user facing this issue.

To fix this, i've updated the logic so that we pick the first root project and all of its dependents.

### DIFF
--- a/src/testModel.ts
+++ b/src/testModel.ts
@@ -118,9 +118,43 @@ export class TestModel extends DisposableBase {
         firstProject = false;
       }
     } else {
-      if (this.projects().length)
-        this.projects()[0][kIsEnabled] = true;
+      const inner = new Set<string>();
+      for (const { project: { dependencies, teardown } } of this._projects.values()) {
+        for (const dep of dependencies)
+          inner.add(dep);
+        if (teardown)
+          inner.add(teardown);
+      }
+
+      const firstRootProject = this.projects().find(p => !inner.has(p.name));
+      if (!firstRootProject)
+        return;
+
+      for (const project of this._dependencies(firstRootProject))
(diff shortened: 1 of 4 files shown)


src/testModel.ts
```diff
@@ -118,11 +118,45 @@ export class TestModel extends DisposableBase {
         firstProject = false;
       }
     } else {
-      if (this.projects().length)
-        this.projects()[0][kIsEnabled] = true;
+      const inner = new Set<string>();
```
dgozman: Let's extract a method `_markInitiallyEnabledProjects()`.

src/testModel.ts
```diff
@@ -118,11 +118,45 @@ export class TestModel extends DisposableBase {
...
+      const firstRootProject = this.projects().find(p => !inner.has(p.name));
+      if (!firstRootProject)
+        return;
+
+      for (const project of this._dependencies(firstRootProject))
+        project[kIsEnabled] = true;
     }
   }
 
+  private _dependencies(project: TestProject) {
```
dgozman:
```suggestion
  private _projectClosure(project: TestProject) {
```

Skn0tt:
Closing this for now to leave a clean table as i'm going OOO. Team discussion results are in https://github.com/microsoft/playwright/issues/35731#issuecomment-3271393869.

PR #692, opened 2025-09-30
feat: enable all project by default, point user towards project selector on first test run

Skn0tt:
implements this approach from https://github.com/microsoft/playwright/issues/35731#issuecomment-3271393869:

> show test triangles for all projects, and on first test run show a modal that prompts user to select projects.

The modal was a little too much, so we went with the gentler approach of revealing the projects selector, and adding a little explanatory notification.

https://github.com/user-attachments/assets/7d510993-60a8-489f-9bd7-f13f3b7371df


tests/run-tests.spec.ts
```diff
@@ -458,76 +458,6 @@ test('should run all projects at once', async ({ activate }) => {
   ]);
 });
 
-test('should group projects by config', async ({ activate }) => {
```
Skn0tt, line 461: this test seems to be a remnant from old times. there's no grouping in the UI anymore, but a dropdown instead. so I don't think it's valuable.

package.json
```diff
@@ -104,6 +104,11 @@
         "category": "Test",
         "command": "pw.extension.command.clearCache",
         "title": "%contributes.command.pw.extension.command.clearCache%"
+      },
+      {
+        "category": "Test",
+        "command": "pw.extension.command.clearWorkspaceSettings",
+        "title": "%contributes.command.pw.extension.command.clearWorkspaceSettings%"
```
dgozman: Why do we need this one? Should this be a separate PR?

media/common.css
```diff
@@ -112,6 +112,58 @@ textarea {
   background: none !important;
 }
 
+.project-notification {
```
dgozman, line 115: Why isn't this in `settingsView.css`?

src/settingsView.ts
```diff
@@ -243,6 +248,12 @@ function htmlForWebview(vscode: vscodeTypes.VSCode, extensionUri: vscodeTypes.Ur
           </div>
         </h2>
         <div data-testid="projects" id="projects" class="vbox"></div>
+        <div id="project-notification" class="project-notification" role="region">
+          <button class="project-notification-close" type="button" title="Close Hint">
+            ${closeIcon}
+          </button>
+          <span class="project-notification-text">By default, all projects are enabled. Uncheck projects you don't want to run.</span>
```
dgozman: l10n please!

src/testModel.ts
```diff
@@ -107,7 +107,7 @@ export class TestModel extends DisposableBase {
     if (!this.isEnabled)
       return;
     await this._listFiles();
-    if (configSettings) {
+    if (configSettings && configSettings.projects.length) {
```
dgozman: Why this change? Do you want to enable all projects for someone who has disabled them all manually?

src/testModel.ts
```diff
@@ -799,6 +799,11 @@ export class TestModelCollection extends DisposableBase {
     return (workspaceSettings.configs || []).find(c => c.relativeConfigFile === path.relative(config.workspaceFolder, config.configFile));
   }
 
+  isFreshOpen(): boolean {
+    const workspaceSettings = this.embedder.context.workspaceState.get(workspaceStateKey) as WorkspaceSettings | undefined;
+    return !workspaceSettings || !workspaceSettings.configs || workspaceSettings.configs.length === 0;
```
dgozman: Should this be per-config instead? Flash again when a new config was added sounds reasonable.

src/extension.ts
```diff
@@ -262,6 +262,13 @@ export class Extension implements RunHooks {
       this._treeItemObserver,
       registerTerminalLinkProvider(this._vscode),
     ];
+    if (this._context.extensionMode === this._vscode.ExtensionMode.Development) {
```
dgozman, line 267: If this is only for our own tests, we usually just call methods on the `extension` instance.

package.json
Skn0tt: It's for manual testing. Since this feature is based on the memento state, it's useful to be able to clear it. There's no good way of clearing this built into VS Code though, so I added a command pallette command that's enabled only in local dev.

media/common.css
Skn0tt: copilot :D I'll move it.

src/extension.ts
Skn0tt: see https://github.com/microsoft/playwright-vscode/pull/692#discussion_r2398300256

src/testModel.ts
Skn0tt: unsure about that. to me, this notification is like a guided tour through the extension. after showing it once, the user will know that the project switcher exists and where it's located. maybe we should have a separate notification that highlights the config switcher, that triggers when multiple configs are detected?

Skn0tt: having said that, maybe these notifications should be dependent on global state, not per-workspace.

Skn0tt:
turning this into draft until I get back to it

media/common.css
Skn0tt: you tricked me - there is no such thing as a `settingsView.css`!

src/testModel.ts
Skn0tt: I don't remember exactly. I think I noticed a case in local development, where we store the config settings but with `projects: []`, and then this logic doesn't properly work. I reverted it for now, but will try to recreate it.

Skn0tt:
I refactored this so that the setting is shown once across all installations of VS Code. This makes the notification more of an "onboarding walk through" kind of thing.

src/extension.ts
dgozman: Let's make the name indicate it's for our internal testing.

tests/utils.ts
```diff
@@ -219,10 +219,15 @@ export async function selectConfig(vscode: VSCode, label: string) {
 
 export async function enableProjects(vscode: VSCode, projects: string[]) {
   const webView = vscode.webViews.get('pw.extension.settingsView')!;
+  for (const project of projects)
+    await expect(webView.getByTestId('projects').locator('div').locator('label').getByLabel(project)).toBeVisible();
```
dgozman:
```suggestion
    await expect(webView.getByTestId('projects').getByLabel(project)).toBeVisible();
```

dgozman (approved):

PR #745, opened 2026-01-13
test: fix raciness under congestion

Skn0tt:


src/utils.ts
```diff
@@ -40,17 +40,19 @@ export function stripBabelFrame(text: string) {
...
-  childProcess.stdout.on('data', data => output += data.toString());
-  return new Promise<string>((f, r) => {
+  let stdout = '';
+  let stderr = '';
+  childProcess.stdout.on('data', data => stdout += data.toString());
+  childProcess.stderr.on('data', data => stderr += data.toString());
+  return new Promise((f, r) => {
     childProcess.on('error', error => r(error));
-    childProcess.on('exit', () => f(output));
+    childProcess.on('close', code => f({ stdout, stderr, code: code }));
```
Skn0tt: this is the key change - `exit` doesn't wait for buffered stdio to be reported, `close` does. Under resource congestion, this race condition makes us return `stdout: ''`. The rest of the PR is just housekeeping.

PR #756, opened 2026-02-13
Record start url support

thombrink:
**Summary**
  - Adds a playwright.recordStartUrl VS Code setting that configures a URL to navigate to when using "Record new"
  - When set, the generated test template includes await page.goto(url) which navigates the browser before recording starts
  - Includes URL scheme validation (http://, https://, file://, etc.) with a localized error message
  - When empty (default), behavior is unchanged — recording starts on about:blank

**Motivation**
Currently, clicking "Record new" opens the browser at about:blank and the user must manually navigate to their application. The CLI equivalent (playwright codegen <url>) supports a URL argument, but the VS Code extension had no equivalent. This setting closes that gap.

**How it works**
The URL is embedded as page.goto() in the generated test template. The test runs before recording starts, navigating the browser in-place — no subprocess or CDP manipulation needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,6 +154,13 @@
           ],
           "default": "3way",
           "description": "%configuration.playwright.updateSourceMethod%"
+        },
+        "playwright.recordStartUrl": {
+          "type": "string",
+          "default": "",
+          "pattern": "^$|^[a-zA-Z]+://",
+          "patternErrorMessage": "%configuration.playwright.recordStartUrl.patternErrorMessage%",
+          "description": "%configuration.playwright.recordStartUrl%"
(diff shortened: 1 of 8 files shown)


thombrink:
@microsoft-github-policy-service agree

thombrink:
It also fixes issue [20059](https://github.com/microsoft/playwright/issues/20059)

Skn0tt:
Thanks for the PR, sorry for the very late review. I dont like that this introduces additional complexity. If you need this for your usecase, I recommend overriding the page fixture and calling goto in there, this should have the same effect.

PR #757, opened 2026-02-17
fix: range end line was still 1-based causing it to bleed into the next line

SivanRP:
Fixes #38911

## What was wrong

I was digging into why Test Explorer arrows were showing up on the wrong lines and found it in `_syncSuite` inside `src/testTree.ts`. Playwright gives us line numbers starting at 1, but VS Code's `Range` API uses 0-based line numbers (line 0 = first line of the file).

The start line was already being converted correctly with `line - 1`, but the end line was still using the raw value from Playwright. So every `TestItem` range was stretching one line further than the actual `test()` call.

## What I changed

**`src/testTree.ts`** — compute a single `zeroBased` value and use it for both start and end of the `Range`:

```ts
// before
vsChild.range = new this._vscode.Range(Math.max(line - 1, 0), 0, line, 0);

// after
const zeroBased = Math.max(uChild.location.line - 1, 0);
vsChild.range = new this._vscode.Range(zeroBased, 0, zeroBased, 0);
```

I also removed the `else if` branch below it — it could never actually execute because the conditions were mutually exclusive.

**`tests/location.spec.ts`** — added a regression test that checks `range.start.line === range.end.line` (which would have failed on the old code) and verifies the start line is the correct 0-based value.

### DIFF
--- a/src/testTree.ts
+++ b/src/testTree.ts
@@ -166,10 +166,12 @@ export class TestTree extends DisposableBase {
         vsChild.tags = uChild.tags.map(tag => new this._vscode.TestTag(tag));
       const hasLocation = uChild.location.line || uChild.location.column;
       if (hasLocation && (!vsChild.range || vsChild.range.start.line + 1 !== uChild.location.line)) {
-        const line = uChild.location.line;
-        vsChild.range = new this._vscode.Range(Math.max(line - 1, 0), 0, line, 0);
-      } else if (hasLocation && !vsChild.range) {
-        vsChild.range = undefined;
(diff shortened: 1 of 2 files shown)


Skn0tt (commented):
Good job spotting the dead code! I wonder how that came to be.

My understanding is that VS Code places the green test triangle at the beginning of the test item range, so fixing the end location shouldn't impact anything.

Can you reproduce the bug from the issue, where the green triangles are misaligned? In my testing, everything looks good.

Skn0tt:
Closing as stale, let's start from a reproducible bug report.

PR #764, opened 2026-03-27
test: cover ESM-only dependency discovery

raashish1601:
Refs microsoft/playwright#39902.

## Summary
- add a VS Code extension integration test for a workspace that uses ESM config and imports a local module that depends on an ESM-only package
- verify test discovery still lists the spec in the extension tree for that setup

## Testing
- `npx eslint tests/list-tests.spec.ts`
- `$env:PLAYWRIGHT_BROWSERS_PATH=''C:\Users\raash\Desktop\open-source-contribution\.tmp-contributor-03-pw-browsers''; npx playwright test tests/list-tests.spec.ts -g "should list tests when they import local modules with ESM-only package dependencies"`

### DIFF
--- a/tests/list-tests.spec.ts
+++ b/tests/list-tests.spec.ts
@@ -83,6 +83,44 @@ test('should list tests for visible editors', async ({ activate }) => {
   ]);
 });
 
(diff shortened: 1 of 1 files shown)


Skn0tt:
test is green, let's figure out what's going on in the issue first

PR #769, opened 2026-04-18
fix(extension): bind spawned servers to 127.0.0.1 instead of localhost

yury-s:
Refs https://github.com/microsoft/playwright/issues/40226 (root-cause fix; companion to #768 which prevents the follow-on crash).

## Summary
Users started hitting `connect ETIMEDOUT ::1:<port>` after Windows 11 **25H2** (build 26200.8246), with matching reports on macOS. The extension spawns Playwright's `test-server` (and, for "reuse browser", `run-server`) without specifying `--host`. Both default to `localhost`, which Node resolves to `::1` first on dual-stack systems; the server binds IPv6-only, and something on 25H2 (firewall/WFP/MpsSvc defaults appear to be the culprit) then blocks the extension's IPv6 loopback connection.

- Pass `--host 127.0.0.1` to [`test-server`](src/playwrightTestServer.ts#L409) (listing + debug paths) and [`run-server`](src/reusedBrowser.ts#L91-L95). IPv4 loopback is reachable on all affected setups.
- No behavior change on systems that were already working — it just pins the bind address instead of relying on resolver order.
- Added [`tests/list-tests.spec.ts`](tests/list-tests.spec.ts) "should spawn test-server bound to IPv4 loopback" that asserts the spawn args via the captured log output channel. Verified the test fails when the `--host` arg is removed.

The existing `#768` defensively fixes the resulting `Position(-1)` crash; this PR removes the trigger.

### DIFF
--- a/src/playwrightTestServer.ts
+++ b/src/playwrightTestServer.ts
@@ -319,7 +319,7 @@ export class PlaywrightTestServer {
           PWDEBUG: 'console',
         },
         program: paths.cli,
-        args: ['test-server', '-c', paths.config],
+        args: ['test-server', '-c', paths.config, '--host', '127.0.0.1'],
       });
 
       if (token?.isCancellationRequested)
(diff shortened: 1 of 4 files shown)


pavelfeldman (approved):

pavelfeldman (changes requested):
stale defender rules problem

pavelfeldman (approved):

PR #789, opened 2026-06-19
fix install command quoting on Windows cmd

mturac:
Fixes microsoft/playwright#40076

## Summary
- quote Install Playwright flags with double quotes on Windows so cmd does not pass literal single quotes
- add regression coverage for the generated terminal command

## Tests
- npx playwright test tests/installer.spec.ts --project=default
- npx eslint src/installer.ts tests/installer.spec.ts
- npx tsc --noEmit
- npm run lint
- npm run build

### DIFF
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -72,6 +72,9 @@ export async function installPlaywright(vscode: vscodeTypes.VSCode) {
 }
 
 function quote(s: string): string {
+  // Windows cmd treats single quotes as literal characters.
+  if (process.platform === 'win32')
+    return `"${s}"`;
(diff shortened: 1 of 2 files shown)


dgozman:
@Skn0tt Could you please take a look?

Skn0tt (commented):
We don't know whether the shell is CMD or Powershell at this point, and Powershell treats double quotes as expandable. I'd rather solve this by detecting the exact shell, or by ensuring no callers pass user-provided data.

Skn0tt:
I've implemented that over in https://github.com/microsoft/playwright-vscode/pull/792

PR #794, opened 2026-07-07
fix: discover config even when hidden by files.exclude

### LINKED ISSUE #41601: [Bug]: VS code extension says: No Playwright tests found.
### Version

1.61.1

### Steps to reproduce

1. Clone my repo at https://github.com/Sbrjt/playwright
2. npm install
3. npx playwright test
4. The tests run from terminal, but the vscode extension says "No Playwright tests found."

If I open in Kiro I can see the tests tho.

### Expected behavior

Works in [Kiro](https://github.com/kirodotdev/Kiro) (based on VS Code 1.107.1)

Image

### Actual behavior

Not working in VS Code

Image

### Additional context

_No response_

### Environment

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (18) x64 Intel(R) Core(TM) Ultra 5 125H
    Memory: 271.54 MB / 15.61 GB
  Binaries:
    Node: 24.13.0 - C:\nvm4w\nodejs\node.EXE
    npm: 11.14.1 - C:\nvm4w\nodejs\npm.CMD
  IDEs:
    VSCode: 1.126.0 - C:\Program Files\Microsoft VS Code\bin\code.CMD
  npmPackages:
    @playwright/test: ^1.61.1 => 1.61.1
```

Skn0tt:
If a user's `files.exclude` matches their Playwright config (e.g. `"files.exclude": { "**/*config*": true }`), the extension reports "No Playwright tests found".

Turns out `findFiles` applies the user's `files.exclude`/`search.exclude` unless you pass `null` as the exclude arg. We were passing a glob string, so the config got hidden. This PR passes `null` instead - the config is machinery we need regardless of what's hidden in the file tree - and filters `node_modules` in the loop instead.

The `findFiles` mock now honours `files.exclude` like real VS Code, otherwise the bug isn't reproducible and the regression test would pass either way.

Also verified manually in VS Code with the failing setting - the tests show up now.

Closes https://github.com/microsoft/playwright/issues/41601

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -310,7 +310,7 @@ export class Extension implements RunHooks {
     this._models.clear();
     this._testTree.startedLoading();
 
-    const configFiles = await this._vscode.workspace.findFiles('**/*playwright*.config.{ts,js,mts,mjs}', '**/node_modules/**', undefined, token);
+    const configFiles = await this._vscode.workspace.findFiles('**/*playwright*.config.{ts,js,mts,mjs}', null, undefined, token);
     if (token.isCancellationRequested)
       return;
 
@@ -319,6 +319,8 @@ export class Extension implements RunHooks {
     for (const configFileUri of configFiles) {
       const configFilePath = uriToPath(configFileUri);
       // TODO: parse .gitignore
+      if (configFilePath.includes('node_modules'))
+        continue;
       if (!this._isUnderTest && configFilePath.includes('test-results'))
         continue;
 
--- a/tests/mock/vscode.ts
+++ b/tests/mock/vscode.ts
@@ -1185,16 +1185,28 @@ export class VSCode {
       return document;
     };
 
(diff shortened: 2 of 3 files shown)


dgozman (commented):
I am not really sure we want to fix this. If config files are hidden (what does that even mean?) - perhaps the user does not want the IDE to pick them up, including the playwright extension? We assume that playwright config is committed to the source tree and is not "hidden".

Skn0tt:
@dgozman We can ask OP, but I think they're excluding `*config*` because they don't like how their sidebar is cluttered with dozens of these files. The default for hidden files includes `.git`, and still users want the Git extension to be enabled even though the relevant files are excluded from the sidebar.

Users that really don't want the Playwright extension to be enabled, can just disable the extension explicitly.

dgozman (approved):
Convinced.

PR #806, opened 2026-08-25
fix: resolve inline source maps during test discovery

### LINKED ISSUE #42364: [Feature]: VS Code extension should support inline source maps during test discovery
### 🚀 Feature Request

The Playwright VS Code extension should resolve inline source maps when discovering test files.

Playwright itself already understands inline source maps. For a generated JavaScript test ending with:

```js
//# sourceMappingURL=data:application/json;charset=utf-8;base64,...
```

Playwright correctly maps tests to their original source files and locations.

However, the VS Code extension's initial file discovery only supports external source-map files. Its current [`resolveSourceMap()` implementation](https://github.com/microsoft/playwright-vscode/blob/afe7e6a247e87561549488444edf9cba97c48524/src/utils.ts#L57-L89):

1. Reads the final `sourceMappingURL`.
2. Resolves its value as a filesystem path.
3. Attempts to read that path as a `.map` file.

For an inline `data:` URL, the file read fails and the extension falls back to the generated JavaScript file. Consequently, the original source file is not registered in Test Explorer and does not receive editor gutter run/debug controls.

The generated JavaScript file is not usable as a fallback in the extension either. The discovery flow produces the following results:
```text
listFiles()                     → generated .spec.js file with zero tests
listTests(generated .spec.js)   → no tests
listTests(original source file) → mapped tests
```

When the extension requests tests for the generated file, Playwright reads the inline source map and treats the mapped original source as the canonical test location. The generated path therefore does not match the file filter, and no tests are returned.

At the same time, the extension never registered the original source file because it could not decode the inline source map during initial discovery. This leaves neither file with test items or gutter controls.

### Proposal

Could `resolveSourceMap()` detect and decode inline source-map data URLs while preserving the existing external source-map behavior?

Inline and external source maps should produce the same Test Explorer hierarchy and editor gutter controls.

The existing [`tests/source-map.spec.ts`](https://github.com/microsoft/playwright-vscode/blob/afe7e6a247e87561549488444edf9cba97c48524/tests/source-map.spec.ts) coverage could be extended with an inline `data:application/json;charset=utf-8;base64,...` source map.

### Motivation

Test generators and compilers may use inline source maps to avoid managing additional generated `.map` files. Because the Playwright runner already supports inline source maps, the difference in VS Code discovery is unexpected.

dcrousso:
inline source maps fall back to generated test files and leave Test Explorer empty

decode base64 `data:` URLs before mapping discovered tests back to their source files

fixes <https://github.com/microsoft/playwright/issues/42364>


yury-s (approved):

vitalets:
Could you share, when do you plan to release?

PR #807, opened 2026-08-29
fix: don't insert stale recorded action after user edits the file

### LINKED ISSUE #42218: [Bug]: Record at cursor inserts stale/previous action after editing test while recording (browser focus loss)
### Version

@playwright/test: 1.62.1, Playwright Test for VS Code extension: 1.1.19, macOS

### Steps to reproduce

1. Enable Show browser in the Playwright Testing sidebar.
2. Run a test so the browser is left open at a useful page state.
3. Place the cursor inside the test where new steps should be inserted.
4. Click Record at cursor.
5. Perform a few actions in the browser — they are inserted correctly at the cursor.
6. Switch focus to the editor and manually edit the test (e.g. tidy a locator, add a comment, delete a generated line, move the cursor).
7. Switch back to the browser without stopping recording.
8. Perform one more click/action in the browser.

### Expected behavior

Only the new action is appended at the current cursor position.

### Actual behavior

The new action is recorded, and an extra stale action is also inserted — typically the last (or penultimate) action from earlier in the same recording session, or sometimes from a previous recording session.

### Workaround
Stop recording before editing the file. After editing, click Record at cursor again and continue. That avoids the stale action, but significantly slows the record → tidy → continue workflow.

### Notes
If I stay entirely in the browser and never edit mid-recording, generation behaves correctly.
The failure is specifically tied to leaving the browser / editing the file while recording remains active, then resuming clicks.
This feels like the recorder’s internal action buffer / document rewrite getting out of sync with the editor buffer (similar class of issue to editing-while-recording problems discussed in #17771).

### Environment

```shell
@playwright/test: 1.62.1
Playwright VS Code extension: 1.1.19
IDE: Cursor (VS Code-compatible)
OS: macOS Darwin 25.x
```

ashrafiucse:
## Summary
- guard the `sourceChanged` re-render path: only update the editor while the selection still spans the exact text the extension previously inserted; when the user edited the file during recording, their edits are left alone instead of re-inserting the re-rendered action at their cursor
- legitimate re-renders (dialog/assertion signals, merged fills) still update the editor while the recorded line is untouched

Fixes https://github.com/microsoft/playwright/issues/42218

The cross-session variant (stale actions from a previous recording session) is fixed core-side: https://github.com/microsoft/playwright/pull/42458

Related: #808 keeps the browser open after stopping a debug session, extending the record-after-debug workflow (https://github.com/microsoft/playwright/issues/37822).


dgozman:
This is very involved, we'd prefer to not do this unless the issue is very common. Thank you for the PR.

PR #808, opened 2026-08-29
fix: keep reused browser when stopping the debugging

ashrafiucse:
## Summary
- stopping a debug session tears down the debugged test process and its page; when the page count then hit zero, `onDidRunTests()` and the late `_pageCountChanged(0)` update both killed the reused browser backend, closing the browser instead of keeping it for inspection and recording
- keep the backend alive after a debug run ends with "Show Browser" enabled, covering both the immediate teardown and the late page-count update

Addresses https://github.com/microsoft/playwright/issues/37822 (the issue is closed by the root-cause fix below; this hardens the extension side and covers workspaces pinned to older cores).

The root cause (pages torn down with the connection) is fixed core-side: https://github.com/microsoft/playwright/pull/42461

Related: #807 fixes stale recorder actions inserted while editing during "Record at cursor" (https://github.com/microsoft/playwright/issues/42218).


ashrafiucse:
Quick context for reviewers: I initially explored a core-side fix for this in https://github.com/microsoft/playwright/pull/42461, but closed it after the feedback there — changing how reuse-mode connections tear down isn't the direction you want. So this PR deliberately stays on the extension side: core behavior is untouched, it just stops the extension from killing the backend when the page count briefly drops to zero after a debug session ends.

I re-verified the fix and its regression test against current playwright main (with #42458 merged in core) and everything still passes. Without this change the issue is still reproducible — and judging by the latest comment on the issue, the suggested workaround doesn't fully work either — so keeping this open for review.

That said, if you'd rather handle it differently (e.g. core-side after all), I'm happy to close or rework this.

pavelfeldman:
Please start with filing an issue!

